PR #780, opened 2026-02-17
fix: stop Socket.IO reconnect thrashing on effect re-runs

### LINKED ISSUE #756: bug: Socket.IO reconnect thrashing — 118 connect/disconnect events in 50 seconds
## Description

A single user session generated **118 socket connect/disconnect events** between 12:44:27 and 12:45:22 (~55 seconds). All three namespaces (monitoring, llm, remediation) disconnect and reconnect **in lockstep every 1-2 seconds**.

This is consistent with a frontend pattern that tears down and recreates socket connections on every route change, or a \`useEffect\` with incorrect dependency arrays that causes socket reconnection on every render.

Each reconnect also triggers a burst of HTTP API calls (harbor/enabled, remediation/actions, settings), causing **"premature close" errors (51 occurrences)** when old requests are aborted mid-flight.

## Steps to Reproduce

1. Open the dashboard
2. Navigate between pages rapidly
3. Monitor backend logs for socket connect/disconnect events
4. Observe: sockets disconnect and reconnect on every navigation

## Expected Behavior

- Socket connections should persist across route changes
- Sockets should only disconnect on page unload or explicit logout

## Actual Behavior

- 118 connect/disconnect events in 50 seconds from a single user
- All 3 namespaces thrash in lockstep
- Each reconnect triggers redundant API calls
- 51 "premature close" errors from aborted in-flight requests

## Root Cause

Socket connection lifecycle is tied to component mount/unmount rather than persisting at the app level. Likely a \`useEffect\` in a socket provider that reconnects when route/component changes.

## Affected Files

- \`frontend/src/\` (socket provider/hooks: socket connection lifecycle needs to be lifted to app-level provider that persists across route changes)

## Suggested Fix

1. Move socket connections to an app-level provider (above the router) that creates connections once on mount
2. Use refs instead of state for socket instances to prevent re-renders triggering reconnections
3. Ensure \`useEffect\` cleanup only disconnects on app unmount, not route changes
4. Consider a connection manager singleton pattern

## Priority

**P1 - High**: Causes massive unnecessary load (118 reconnections + cascading API calls), "premature close" errors, and poor user experience. Directly contributes to backend instability.

## Environment

- Project: ai-portainer-dashboard v2.0.0
- Branch: \`dev\`

laboef1900:
## Summary

Closes #756

Two root causes fixed:

**1. socket-provider.tsx: cleanup disconnecting all sockets on every effect re-run**

The cleanup function called `socket.disconnect()` on every effect cleanup — which runs any time `isAuthenticated`, `token`, or `potatoMode` changes. This tore down all three WebSocket connections simultaneously, then the new effect run created fresh ones, racing with socket.io's own reconnection timers.

Fix: cleanup only removes event listeners (`s.off(...)`). Sockets are only explicitly disconnected on logout (`isAuthenticated` becomes false via `disconnectAll()`). Socket.io handles reconnection automatically.

**2. socket.ts: getNamespaceSocket() replacing in-flight reconnecting sockets**

`getNamespaceSocket()` checked `existing?.connected` and deleted disconnected sockets from cache, immediately creating new ones. When a socket was in the middle of its reconnection backoff, this created a second socket racing with the first, and reset the backoff timer on every cycle.

Fix: return the existing socket regardless of connection state — it will reconnect on its own. Only evict from cache on `'io client disconnect'` (explicit client-initiated), not on `'transport close'` or `'io server disconnect'` (transient errors). Added `reconnectionDelayMax: 30_000` and `randomizationFactor: 0.5` for proper exponential backoff with jitter.

## Test plan

- [x] `cd frontend && npx vitest run src/lib/socket.test.ts src/providers/socket-provider.test.tsx` — 16 tests pass
- [x] `npm run typecheck -w frontend` — no type errors
- New test: "reuses disconnected socket for same namespace+token (avoids thrashing)"
- New test: "does not disconnect sockets when effect re-runs (no thrashing)"
- New test: "calls disconnectAll on logout"

### DIFF
--- a/frontend/src/lib/socket.test.ts
+++ b/frontend/src/lib/socket.test.ts
@@ -44,6 +44,13 @@ describe('socket singleton', () => {
       expect(mockIo).toHaveBeenCalledTimes(1);
       expect(socket).toBeDefined();
     });
+
+    it('reuses existing main socket on subsequent calls', () => {
+      const socket1 = getSocket('token-1');
+      const socket2 = getSocket('token-1');
+      expect(socket1).toBe(socket2);
+      expect(mockIo).toHaveBeenCalledTimes(1);
+    });
   });
 
   describe('getNamespaceSocket', () => {
@@ -55,13 +62,22 @@ describe('socket singleton', () => {
 
     it('reuses connected socket for same namespace+token', () => {
       const socket1 = getNamespaceSocket('monitoring', 'token-1');
-      // Simulate connected state
       (socket1 as any).connected = true;
 
       const socket2 = getNamespaceSocket('monitoring', 'token-1');
-      // Should return same socket, not create new one
       expect(socket2).toBe(socket1);
-      // io should only have been called once
+      expect(mockIo).toHaveBeenCalledTimes(1);
+    });
+
+    it('reuses disconnected socket for same namespace+token (avoids thrashing)', () => {
+      // Socket.io reconnects automatically — we must not replace it with a new
+      // instance just because it is temporarily disconnected, otherwise we race
+      // with its built-in reconnection timer.
+      const socket1 = getNamespaceSocket('monitoring', 'token-1');
+      (socket1 as any).connected = false;
+
+      const socket2 = getNamespaceSocket('monitoring', 'token-1');
+      expect(socket2).toBe(socket1);
       expect(mockIo).toHaveBeenCalledTimes(1);
     });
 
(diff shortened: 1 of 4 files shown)


laboef1900:
## Automated PR Review

**Verdict: COMMENT** :mag:

### Summary
Excellent fix for Socket.IO reconnect thrashing. The two root causes are correctly identified and surgically fixed: (1) cleanup disconnecting all sockets on every React effect re-run, and (2) replacing in-flight reconnecting sockets. The fix is minimal and well-tested with 3 new tests. CI is fully green.

### Findings

| # | Severity | File | Line(s) | Finding |
|---|----------|------|---------|---------|
| 1 | INFO | `socket.ts` | 17 | **`getSocket` no longer checks `connected` state**: Previously returned cached socket only when connected. Now returns it regardless. This means if the main socket's token is stale (e.g., after a token refresh), it will continue using the old socket. However, since the socket's `auth` option is set at creation time and socket.io sends the auth on reconnect, this should be fine as long as the token doesn't change. If the token does change, `disconnectAll()` is called on logout which clears the cache. **No fix needed.** |
| 2 | INFO | `socket-provider.tsx` | 48-55 | **Splitting connected state into separate `useState`**: The `connected` boolean is now a separate state from the socket references. This avoids unnecessary re-renders from the socket objects themselves when only the connection status changes. Good optimization. **No fix needed.** |

### Fixes Applied
None required -- the fix is correct, minimal, and well-tested.

---
:robot: Reviewed by PR Reviewer Agent

PR #787, opened 2026-02-18
feat: implement secure coding paradigms for LLM data and tool use

laboef1900:
Introduces three core security paradigms for LLM interactions:
1. Privacy: New PII Scrubber utility masks emails, IPs, UUIDs, and secrets.
   Integrated into llm-client and container log tools to prevent PII leakage.
2. Robustness: Replaces naive JSON parsing with Zod-validated tool calling
   to handle non-deterministic LLM output with safe fallbacks.
3. Sandboxing: Adds 'requiresApproval' flag to tool definitions.
   The tool executor now blocks unapproved execution of sensitive tools.

Includes:
- backend/src/utils/pii-scrubber.ts (New)
- backend/src/services/llm-client.ts (PII integration)
- backend/src/services/llm-tools.ts (Zod + Approval logic)

### DIFF
--- a/backend/src/services/llm-tools.ts
+++ b/backend/src/services/llm-tools.ts
@@ -22,7 +22,10 @@ const ToolCallSchema = z.object({
     name: z.string(),
     arguments: z.union([z.string(), z.record(z.unknown())]).optional(),
   }).optional(),
-});
+}).refine(
+  (data) => data.tool !== undefined || data.function !== undefined,
+  { message: 'Tool call must have either "tool" or "function" field' },
+);
 
 /** Check if a tool name is an MCP-prefixed name (e.g. mcp__kali-mcp__run_allowed). */
 function isMcpToolName(name: string): boolean {
--- a/backend/src/utils/pii-scrubber.ts
+++ b/backend/src/utils/pii-scrubber.ts
@@ -65,9 +65,18 @@ export function scrubPii(text: string, options: ScrubberOptions = {}): string {
   // 2. Secret assignments (handling capturing groups to keep the key)
   scrubbed = scrubbed.replace(PII_PATTERNS.secretAssignment, (match, p1) => {
     matchCount++;
-    // We want to keep the "key=" part and only mask the value
-    const keyPart = match.substring(0, match.indexOf(p1));
-    const suffix = match.substring(match.indexOf(p1) + p1.length);
+    // We want to keep the "key=" part and only mask the value.
+    // Defensive: if the captured value is not found (e.g. earlier patterns already
+    // scrubbed part of the match), fall back to replacing the entire match.
+    const idx = match.indexOf(p1);
+    if (idx === -1) {
+      if (options.verbose) {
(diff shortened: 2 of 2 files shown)


laboef1900:
## Automated PR Review

**Verdict: COMMENT** :mag:

### Summary
Solid security-oriented PR introducing PII scrubbing, Zod-validated tool calls, and a requiresApproval gate. Well-tested with 155 lines of new tests. CI is green. A few minor issues found below.

### Findings

| # | Severity | File | Line | Finding |
|---|----------|------|------|---------|
| 1 | WARNING | `pii-scrubber.ts` | 19 | **Regex uses global flag with stateful `lastIndex`**: The `PII_PATTERNS` regexes use the `/g` flag. When used with `String.prototype.replace()` this is fine, but if these patterns are ever used with `.test()` or `.exec()` in the future, the shared global-flag regex will retain `lastIndex` between calls and alternate between matching and not matching. The current usage via `replace()` is safe, but adding a comment would prevent future misuse. |
| 2 | WARNING | `pii-scrubber.ts` | 66 | **`secretAssignment` regex runs on already-scrubbed text**: The secret assignment pattern runs after emails, API tokens, and JWTs have already been replaced with `[MASKED]`. If a secret value contained an email or JWT, the `p1` capture group position lookup via `match.indexOf(p1)` could return -1 or an incorrect position since the original text was already modified. In practice this is unlikely since secret values rarely contain emails, but defensively checking `indexOf` would be safer. |
| 3 | INFO | `pii-scrubber.ts` | 27 | **`secretAssignment` regex character class has unescaped `]`**: The character class `[a-zA-Z0-9\!@#$%^&*()_+={}\[\]|:;"'<>,.?/~\`-]{8,}` has `[` and `]` inside a character class. While the `[` is escaped, the behavior depends on the JS engine's error recovery. This works in V8 but could be fragile. |
| 4 | INFO | `llm-tools.ts` | 17-25 | **Zod schema allows both `tool` and `function` to be optional**: The `ToolCallSchema` has both `tool` and `function` as optional fields. A completely empty object `{}` would pass validation, which is then caught by the `\!toolName` check later. Not a bug but the schema could be tightened with a `.refine()` to require at least one. |
| 5 | INFO | `llm-tools.ts` | 883 | **`rawArgs` typed as `any`**: The `rawArgs` variable is typed as `any` instead of `Record<string, unknown>`. Minor type safety issue. |
| 6 | INFO | `llm-client.ts` | 184-186 | **PII scrubbing applied to all messages including system prompt**: The system prompt is scrubbed twice -- once via `scrubPii(systemPrompt)` at line 186 and again inside `buildInfrastructureContext` at line 375. Double scrubbing is safe but wasteful. Not a bug. |

### Fixes Applied
- #2: Add defensive `indexOf` check in `secretAssignment` handler to avoid incorrect substring on already-scrubbed text
- #4: Add `.refine()` to ToolCallSchema requiring at least `tool` or `function` to be present

---
:robot: Reviewed by PR Reviewer Agent

PR #1164, opened 2026-05-05
test: add regression test for AuthProvider.logout role reset

### LINKED ISSUE #957: bug: AuthProvider.logout() does not reset role state — stale admin role visible after logout

laboef1900:
## Summary

Issue #957 reports that `AuthProvider.logout()` does not reset the role state to `'viewer'` on sign-out, leaving an admin's privileges visible client-side. **The fix is already on `dev`** — `frontend/src/providers/auth-provider.tsx` line 122 calls `setRole('viewer')` and line 123 calls `clearStoredAuth()`. This PR therefore ships a regression test only and closes the issue.

The test renders a real `<AuthProvider>` with a small consumer, mocks only the HTTP boundary (`api.post` / `api.setToken`) per the project Testing & Mocks rules, drives a login-as-admin → logout flow, and asserts:

- `role === 'viewer'` after logout
- `token === null` and `isAuthenticated === false`
- `localStorage` (`AUTH_TOKEN_KEY`, `auth_username`, `auth_role`) is wiped — per critic recommendation E2, this future-proofs the suite against the `auth:expired` handler diverging from `logout` later.

A second case re-runs the assertions when `/api/auth/logout` rejects, ensuring backend failure cannot strand the user in an admin session client-side.

## Test plan
- [x] `cd frontend && npm run lint` — clean (0 warnings)
- [x] `cd frontend && npm run typecheck` — clean
- [x] `npx vitest run src/providers/auth-provider.test.tsx` — 2/2 pass
- [x] `npx vitest run` (full frontend suite) — 1613/1613 pass, no regressions
- [x] No new test infrastructure introduced; reuses existing `vi.mock('@/shared/lib/api', ...)` pattern.

Closes #957

### DIFF
--- a/frontend/src/providers/auth-provider.test.tsx
+++ b/frontend/src/providers/auth-provider.test.tsx
@@ -1,0 +1,164 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { act, render } from '@testing-library/react';
+import { AUTH_TOKEN_KEY } from '@/shared/lib/auth-constants';
+import { AuthProvider, useAuth } from './auth-provider';
+
+const mockPost = vi.fn();
+const mockSetToken = vi.fn();
+
+vi.mock('@/shared/lib/api', () => ({
+  api: {
+    post: (...args: unknown[]) => mockPost(...args),
+    setToken: (...args: unknown[]) => mockSetToken(...args),
+    getToken: () => null,
+  },
+}));
+
+const AUTH_USERNAME_KEY = 'auth_username';
+const AUTH_ROLE_KEY = 'auth_role';
+
+/**
+ * Build a JWT-shaped token (`header.payload.signature`) whose payload has the
+ * given role and an `exp` claim 1 hour in the future. The auth provider only
+ * inspects the payload via base64-decoding the second segment, so a real
+ * signature is unnecessary.
+ */
+function makeJwt(role: 'admin' | 'operator' | 'viewer'): string {
+  const header = btoa(JSON.stringify({ alg: 'HS256', typ: 'JWT' }))
+    .replace(/=+$/, '')
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_');
+  const payload = btoa(
+    JSON.stringify({ role, exp: Math.floor(Date.now() / 1000) + 3600 }),
+  )
+    .replace(/=+$/, '')
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_');
+  return `${header}.${payload}.signature`;
+}
+
+/**
+ * Test consumer that exposes the auth context state through DOM attributes so
+ * tests can assert on it without re-implementing the provider's hook plumbing.
+ */
+function AuthConsumer() {
+  const { role, token, isAuthenticated, login, loginWithToken, logout } = useAuth();
+  return (
+    <div
+      data-testid="auth-state"
+      data-role={role}
+      data-token={token ?? ''}
+      data-authenticated={isAuthenticated ? 'true' : 'false'}
+    >
+      <button
+        data-testid="login-with-token"
+        onClick={() => loginWithToken(makeJwt('admin'), 'alice', 'admin')}
+      >
+        login-with-token
+      </button>
+      <button
+        data-testid="login"
+        onClick={() => {
+          void login('alice', 'pw');
+        }}
+      >
+        login
+      </button>
+      <button
+        data-testid="logout"
+        onClick={() => {
+          void logout();
+        }}
+      >
+        logout
+      </button>
+    </div>
+  );
+}
+
+describe('AuthProvider — logout (regression for #957)', () => {
+  beforeEach(() => {
+    mockPost.mockReset();
+    mockSetToken.mockReset();
+    window.localStorage.clear();
+  });
+
+  afterEach(() => {
+    window.localStorage.clear();
+    vi.clearAllMocks();
+  });
+
+  it('resets role to "viewer", clears token, and wipes localStorage after logout from admin', async () => {
+    // Logout posts to /api/auth/logout — make it succeed so we exercise the
+    // happy path, but the test should also pass if it errors (see next test).
+    mockPost.mockResolvedValue(undefined);
+
+    const { getByTestId } = render(
+      <AuthProvider>
+        <AuthConsumer />
+      </AuthProvider>,
+    );
+
+    // Establish an authenticated admin session.
+    await act(async () => {
+      getByTestId('login-with-token').click();
+    });
+
+    const stateAfterLogin = getByTestId('auth-state');
+    expect(stateAfterLogin.dataset.role).toBe('admin');
+    expect(stateAfterLogin.dataset.authenticated).toBe('true');
+    expect(stateAfterLogin.dataset.token).not.toBe('');
+    expect(window.localStorage.getItem(AUTH_TOKEN_KEY)).not.toBeNull();
+    expect(window.localStorage.getItem(AUTH_USERNAME_KEY)).toBe('alice');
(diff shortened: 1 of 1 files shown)


laboef1900 (commented):
## Automated PR Review

**PR**: #1164 — test: add regression test for AuthProvider.logout role reset
**Branch**: feature/957-auth-logout-test -> dev
**Diff size**: 164 lines (+164/-0) — single new test file
**Reviewed at**: 2026-05-05 17:57 UTC

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | frontend tsc --noEmit clean |
| Lint | PASS | frontend eslint --max-warnings=0 clean |
| New test file | PASS | 2/2 pass (`npx vitest run src/providers/auth-provider.test.tsx`) |
| Bug-already-fixed claim | VERIFIED | `git show dev:frontend/src/providers/auth-provider.tsx` line 122 has `setRole('viewer')`. The fix is on dev. |

### Regression-Test Effectiveness Check

I sanity-checked that the test would have caught the original bug by temporarily replacing both `setRole('viewer')` calls in `auth-provider.tsx` (logout path + auth:expired handler) with no-ops. **Both new test cases failed** with `expected 'admin' to be 'viewer'` — confirming the test genuinely guards #957 and not just structurally similar code.

### Code Review Findings

| — | — | — | No findings. Test follows project rules: mocks at API boundary (`@/shared/lib/api`), renders real `<AuthProvider>` with a small consumer, no internal-component mocking, no pure-utility mocking. Both happy-path (logout succeeds) and sad-path (`/api/auth/logout` rejects) are covered, and the localStorage assertions also future-proof against the auth:expired handler diverging from logout (the E2 critic recommendation cited in the PR body). |

### Acceptance Criteria (Issue #957)

| AC-1 | After logout, `role` returned by `useAuth()` is `'viewer'` | PASS | auth-provider.test.tsx:122 (`expect(stateAfterLogout.dataset.role).toBe('viewer')`) |
| AC-2 | After logout, `token` is `null` and `isAuthenticated` is `false` | PASS | :123-124 |
| AC-3 | localStorage is wiped (token, username, role keys) | PASS | :128-130 |
| AC-4 | API client is told to drop the token | PASS | :133 (`expect(mockSetToken).toHaveBeenCalledWith(null)`) |
| AC-5 | Failure of `/api/auth/logout` does not strand admin session client-side | PASS | :138-163 (second test, mockPost rejects, all assertions still pass) |

### Story Validation (Issue #957)

**Issue intent**: Stale admin role visible after logout because `logout()` did not reset `role` to `'viewer'`. Fix the asymmetry between `logout` and the `auth:expired` handler.
**Story completion**: FULLY ADDRESSED — the bug is fixed on `dev` (auth-provider.tsx:122) and this PR locks it in with a regression test that fails when the fix is reverted (verified during this review).
**User journey**: Before: viewer who was previously admin briefly sees admin-only UI controls after `logout()` because `role` state still says `'admin'`. After: `role` resets to `'viewer'` synchronously inside `logout()`, even when the backend `/api/auth/logout` POST fails.
**Gaps**: None. The PR scope is exactly "regression test only" as the body states.
**Evidence**: auth-provider.test.tsx:91-136 (happy path) + :138-163 (failure path); `git show dev:frontend/src/providers/auth-provider.tsx` line 122 confirms the fix shipped on dev (no source change in this PR).

### Suggested Actions

No actions required. PR is ready to merge.

### Verdict: APPROVE

**Reason**: Bug is already fixed on `dev` (line 122 of auth-provider.tsx); this PR delivers the missing regression test. The test follows project Testing & Mocks rules (mock at HTTP boundary only, render real provider, no internal-component mocking), covers both success and failure of `/api/auth/logout`, and asserts the full post-logout contract (role + token + isAuthenticated + localStorage + api.setToken(null)). I verified the test catches the regression by temporarily reverting the fix — both cases fail with the expected `'admin' vs 'viewer'` assertion error. Lint and typecheck pass.

---
Reviewed by PR Reviewer Agent | 2026-05-05

PR #1228, opened 2026-05-14
fix(ebpf): surface real errors and pre-flight Edge Agent tunnel on Beyla deploy

laboef1900:
## Summary

Deploys to Edge Agent Standard endpoints failed with an opaque "Internal Server Error" toast that masked the underlying Portainer error and gave no path to recovery. This PR addresses four layered causes that turned a known transient condition (Edge Agent's reverse tunnel not yet open) into a dead-end UX.

- **Frontend api client** (`frontend/src/shared/lib/api.ts`) now prefers Fastify's `body.message` over the generic `body.error` class name, so backend errors stop being flattened to "Internal Server Error" at the boundary.
- **`portainerFetchInner`** reads the response body on non-OK responses and includes it in `PortainerError`, exposing causes like "Unable to get the active tunnel".
- **`isPortainerFailure`** excludes the tunnel-not-active condition from the circuit breaker so a sleepy Edge Agent can't open the breaker for an entire endpoint and lock out retries for 30s.
- **`deployBeyla`** pre-flights the tunnel for type-4 endpoints:
  - Fails fast when the agent is offline: `LastCheckInDate` stale > 3× `EdgeCheckinInterval`, or never checked in.
  - Otherwise polls `/api/endpoints/:id/docker/_ping` with a budget of `2 × interval + 15s` (capped at 5 min).
  - The deploy route surfaces these as 503s with actionable messages.
  - Frontend mutation timeout goes from 60s → 120s to fit the tunnel wait plus image pull.

## Test plan

- [x] `frontend/src/shared/lib/api.test.ts` — new cases verify Fastify `body.message` is preferred and `body.error` is still used as a fallback.
- [x] `packages/security/src/__tests__/ebpf-coverage.test.ts` — new cases cover: pre-flight invoked for type-4, budget scales with `EdgeCheckinInterval`, fail-fast for never-checked-in and stale-check-in agents, error path when tunnel never opens, pre-flight skipped for type-1/2.
- [x] `packages/security/src/__tests__/ebpf-coverage-route.test.ts` — new case verifies 503 + actionable message when the deploy hits the tunnel error.
- [x] Full pre-push test suite: 1830 tests pass across all workspaces.

## Behavior before vs after

Before: click Deploy → spinner → "Deploy failed — Internal Server Error" (5–60s). No information, no path forward.

After (against an online but cold Edge Agent): click Deploy → spinner for the tunnel to open (≤ `2 × EdgeCheckinInterval + 15s`) → success.

After (against an offline Edge Agent, e.g. `srv-edge-01` in this incident): click Deploy → fails immediately → "Edge Agent 'srv-edge-01' has not checked in for 148300s (interval 30s). The agent appears offline — verify it is running and can reach Portainer."

### DIFF
--- a/frontend/src/features/security/hooks/use-ebpf-coverage.ts
+++ b/frontend/src/features/security/hooks/use-ebpf-coverage.ts
@@ -136,7 +136,9 @@ export function useDeployBeyla() {
       api.post(
         `/api/ebpf/deploy/${endpointId}`,
         otlpEndpoint ? { otlpEndpoint } : {},
-        { timeoutMs: 60000 },
+        // Edge Agent Standard endpoints wait up to ~45s for the tunnel to open,
+        // then pull the Beyla image (can be 10-30s on first run); allow 120s total.
+        { timeoutMs: 120000 },
       ),
     onSuccess: (data) => {
       queryClient.invalidateQueries({ queryKey: ['ebpf', 'coverage'] });
--- a/frontend/src/shared/lib/api.test.ts
+++ b/frontend/src/shared/lib/api.test.ts
@@ -224,6 +224,33 @@ describe('ApiClient', () => {
       await expect(api.get('/api/broken')).rejects.toThrow('Internal Server Error');
     });
 
+    it('should prefer descriptive message over Fastify generic error class', async () => {
+      // Fastify default error shape: { statusCode, error: 'Internal Server Error', message: 'actual cause' }
+      // Without this, eBPF deploy failures surface only as "Internal Server Error" instead of the underlying Portainer error.
+      mockFetch.mockResolvedValueOnce({
+        ok: false,
+        status: 500,
+        json: () => Promise.resolve({
+          statusCode: 500,
+          error: 'Internal Server Error',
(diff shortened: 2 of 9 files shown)


laboef1900:
Addressed the info-level suggestion from the review in `48f40a85`:

**Extracted `isEdgeTunnelNotActive` helper** (`packages/core/src/portainer/portainer-client.ts`). The three call sites that string-matched Portainer's HTTP 500 body inline (`isPortainerFailure`, `waitForEdgeTunnel`, eBPF deploy route's catch) now route through one named export.

Added a **snapshot test** in `portainer-client.test.ts` that asserts the exact substring (`unable to get the active tunnel`). If Portainer changes the wording in a future release, this is the canary — without it the predicate would silently start returning false, the breaker would trip on cold Edge Agents, the wait loop would give up immediately, and the 503 path would stop catching the condition, all with zero test failures.

The helper also accepts plain strings (the shape `pingEndpointDocker` returns via `result.error`) and returns `false` safely for nullish / non-Error inputs. Existing 28 route tests + 47 service tests still green; +6 new helper tests.

PR #1229, opened 2026-05-14
security: harden Host-header trust + remove reflected user input in error responses

### LINKED ISSUE #1226: nginx: client-controlled Host header forwarded to backend (Host header injection)
## Summary

`frontend/nginx.conf` forwards the client-controlled `$host` variable as the upstream `Host` header on all backend proxy locations. An attacker who can craft the `Host` request header may influence backend behavior whenever the backend uses `request.hostname` / `Host` for URL generation, redirects, password-reset links, OIDC redirects, cache keys, or absolute self-references. Surfaced by Semgrep (`generic.nginx.security.request-host-used`).

## Root Cause

In `frontend/nginx.conf` the proxy blocks pass through the attacker-controlled `Host` value verbatim:

```nginx
# line 81 (SSE log stream)
proxy_set_header Host $host;

# line 94 (/api/)
proxy_set_header Host $host;

# line 107 (/health/)
proxy_set_header Host $host;

# line 114 (/docs)
proxy_set_header Host $host;

# line 125 (/socket.io/)
proxy_set_header Host $host;
```

`$host` resolves from the request's `Host` header (or `:authority` for HTTP/2). When no `server_name` allowlist matches the request, nginx still proxies, so the value is fully attacker-controlled — `curl -H 'Host: evil.example' https://dashboard/api/...` will reach the backend with `Host: evil.example`.

If the Fastify backend ever constructs URLs from `request.hostname`, e.g. in OIDC redirect computation, e-mail templates, audit log entries, or absolute callback URLs, the attacker can poison those values (web-cache poisoning, password-reset link rewriting, OIDC redirect confusion).

Related CWE: [CWE-20: Improper Input Validation](https://cwe.mitre.org/data/definitions/20.html).
Reference: https://sg.run/4x3Z

## Steps to Reproduce

1. Deploy the dashboard behind nginx as configured.
2. Send `curl -H 'Host: attacker.example' https://<dashboard>/api/auth/oidc/start` (or any backend route that emits absolute URLs).
3. Observe: backend may use `attacker.example` for generated URLs.

## Expected Behavior

The upstream `Host` header should be a trusted, constrained value (either the configured backend hostname or a vetted allowlist of public hostnames).

## Fix Approach

Pick one — recommended is option A:

**Option A (simplest)** — pin the upstream Host so the backend always sees a fixed hostname:

```nginx
proxy_set_header Host backend;
```

Apply to every `location` block in `frontend/nginx.conf`. The backend's `request.hostname` will then be the static internal name, and any code that needs the public hostname must read it from `X-Forwarded-Host` (set by the proxy from a vetted value) or from configuration (e.g. `PUBLIC_BASE_URL` env var).

**Option B** — restrict via `server_name` + a map-based allowlist:

```nginx
map $host $allowed_host {
    dashboard.example.com  dashboard.example.com;
    default                dashboard.example.com;  # fallback
}
server {
    server_name dashboard.example.com;
    ...
    proxy_set_header Host $allowed_host;
}
```

Either fix MUST be paired with verifying that backend code does not silently switch on `request.hostname`/`request.headers.host`. Grep for usages and replace with a configured public URL where appropriate.

## Relevant Files

- `frontend/nginx.conf` — lines 81, 94, 107, 114, 125 (5 occurrences)
- Backend code that consumes `request.hostname` / `request.headers.host` (audit pending)

## Acceptance Criteria

- [ ] `frontend/nginx.conf` no longer forwards client-controlled `$host` to the backend on any `location` block.
- [ ] Backend code is audited for `request.hostname` / `headers.host` consumption; any absolute-URL construction sources from configuration, not request headers.
- [ ] Regression test added in `backend/src/routes/security-regression-headers.test.ts` (or a new `security-regression-host-header.test.ts`) asserting that a crafted `Host` / `X-Forwarded-Host` header cannot influence generated absolute URLs (OIDC redirect, password reset, audit logs, etc.).
- [ ] Semgrep `request-host-used` no longer reports on `frontend/nginx.conf`.

> Surfaced by Semgrep 1.159.0 (`p/security-audit` ruleset). H2C-smuggling finding on lines 122-124 was reviewed and confirmed as a **false positive** — the file already implements the documented `map $http_upgrade $connection_upgrade` mitigation (see comment at line 48-53).

### LINKED ISSUE #1227: Reflect untrusted user input in JSON error messages (Fastify routes)
## Summary

Several Fastify route handlers reflect raw user-supplied strings (`body.feature`, `params.feature`, `body.events[*]`) directly into JSON error bodies. While Fastify serializes responses as `application/json` — which prevents direct script execution in modern browsers — echoing untrusted input in error messages is a defense-in-depth issue: any client that later renders the `error` field via `dangerouslySetInnerHTML` / `innerHTML` would re-introduce reflected XSS, and the reflection makes the API a convenient oracle for content-sniffing or log-injection attempts. Surfaced by Semgrep (`javascript.express.web.tainted-direct-response-express`).

## Root Cause

The validation logic emits the rejected value verbatim instead of returning a constant message. Three files, six occurrences:

In `packages/ai-intelligence/src/routes/llm-feedback.ts`:

```typescript
// line 101 — POST /api/llm/feedback
if (!isValidFeature(body.feature)) {
  return reply.code(400).send({ error: `Invalid feature: ${body.feature}` });
}

// line 240 — POST /api/llm/feedback/generate-suggestion
if (!isValidFeature(feature)) {
  return reply.code(400).send({ error: `Invalid feature: ${feature}` });
}
```

In `packages/foundation/src/routes/settings.ts`:

```typescript
// line 313 — GET /api/settings/prompts/:feature/history
if (!PROMPT_FEATURES.some((f) => f.key === feature)) {
  return (reply as any).code(404).send({ error: `Unknown feature: ${feature}` });
}

// line 336 — POST /api/settings/prompts/:feature/rollback
if (!PROMPT_FEATURES.some((f) => f.key === feature)) {
  return (reply as any).code(404).send({ error: `Unknown feature: ${feature}` });
}
```

In `packages/operations/src/routes/webhooks.ts`:

```typescript
// line 80 — POST /api/webhooks
if (!VALID_EVENT_TYPES.includes(evt) && !evt.endsWith('.*')) {
  return reply.status(400).send({ error: `Invalid event type: ${evt}` });
}

// line 132 — PATCH /api/webhooks/:id (same pattern)
```

Related CWE: [CWE-79: Improper Neutralization of Input During Web Page Generation](https://cwe.mitre.org/data/definitions/79.html).
Reference: https://sg.run/zdElz

### Risk assessment

- **Direct XSS risk**: Low. Fastify sets `Content-Type: application/json`, and `reply.send({ error })` JSON-encodes the value.
- **Defense-in-depth risk**: Real. Any future client (notification toast, admin UI, log viewer) that injects `error` into the DOM without escaping re-opens reflected XSS. Frontend toast/error renderers should be audited but are also out of scope for this issue.
- **Log/oracle risk**: Real but minor — the API confirms back the exact rejected value to the caller.

## Expected Behavior

Validation failures should return a static error message; the rejected value, if needed for the caller, belongs in a separate, structured field (e.g. `code: 'invalid_feature'`).

## Fix Approach

Replace each interpolated message with a static one and add a structured code:

```typescript
return reply.code(400).send({
  error: 'Invalid feature',
  code: 'invalid_feature',
});
```

Or, if the client needs to know which value was rejected, return it as a separate field — clients should never display server-supplied error strings unescaped:

```typescript
return reply.code(400).send({
  error: 'Invalid feature',
  code: 'invalid_feature',
  rejectedValue: body.feature,   // structured, never interpolated
});
```

Apply consistently to all 6 sites.

## Relevant Files

- `packages/ai-intelligence/src/routes/llm-feedback.ts` — lines 101, 240
- `packages/foundation/src/routes/settings.ts` — lines 313, 336
- `packages/operations/src/routes/webhooks.ts` — lines 80, 132

## Acceptance Criteria

- [ ] All 6 sites return a static error string; no user-controlled value is interpolated into the `error` field.
- [ ] If the rejected value is still surfaced, it lives in a structured `rejectedValue` / `code` field — not in `error`.
- [ ] Regression test added in `backend/src/routes/security-regression-headers.test.ts` (or a new `security-regression-error-reflection.test.ts`) asserting that a crafted feature/event value (e.g. `<script>alert(1)</script>`) does NOT appear inside the `error` field of the response.
- [ ] Semgrep `tainted-direct-response-express` no longer reports on these files.

> Surfaced by Semgrep 1.159.0 (`p/security-audit` ruleset). Severity is intentionally `priority/low` because Fastify's JSON content-type mitigates direct XSS — this is a hardening / hygiene issue, not an exploitable bug today.

laboef1900:
## Summary

- Closes #1226 — nginx forwarded client-controlled `$host` to backend; trace store could be poisoned via `request.hostname` / `Host` header reads in `request-tracing.ts`.
- Closes #1227 — Eight Fastify route sites reflected user-supplied `feature` / `event` values back into JSON `error` messages (defense-in-depth XSS / log-oracle hardening).

Both findings surfaced by Semgrep (`p/security-audit + p/typescript + p/javascript + p/owasp-top-ten + p/nodejs + p/react`). **After this PR, Semgrep reports 0 findings** on `backend/`, `frontend/`, `packages/` (was 12).

## Changes

**#1226 — Host header injection**
- `frontend/nginx.conf` — `proxy_set_header Host backend` on all 5 `location` blocks (was `$host`). Backend code that needs the public URL must now read it from `DASHBOARD_EXTERNAL_URL`, not from the Host header.
- `packages/core/src/plugins/request-tracing.ts` — Removed `request.hostname` / `request.headers.host` reads. Span attributes no longer embed client-controlled hostnames; `service_name: 'api-gateway'` and the path are enough for trace correlation. `server_address` / `url_full` set to `null`.
- Pre-existing H2C smuggling false positive (lines 128-130) suppressed inline with `# nosemgrep` — the `map $http_upgrade $connection_upgrade` block at lines 50-53 is the documented mitigation and is already covered by a regression test in `security-regression-headers.test.ts:127-140`.

**#1227 — Reflected input in JSON error messages**
- Replaced `\`Invalid feature: ${x}\`` style messages with static error string + structured `code` field, at 8 sites in 4 files:
  - `packages/ai-intelligence/src/routes/llm-feedback.ts` (×2 — Semgrep)
  - `packages/ai-intelligence/src/routes/llm.ts` (×2 — caught during cleanup; not flagged by Semgrep because the return shape was `{success:false, error}` instead of `reply.send(...)`)
  - `packages/foundation/src/routes/settings.ts` (×2 — Semgrep)
  - `packages/operations/src/routes/webhooks.ts` (×2 — Semgrep)
- New shape: `{ error: 'Invalid feature', code: 'invalid_feature' }`. Clients that need the rejected value should add a structured `rejectedValue` field — never display server-supplied `error` strings unescaped.

**Regression tests**
- `backend/src/routes/security-regression-error-reflection.test.ts` (new) — static source-level guard across all 4 files + runtime injection asserting crafted feature/event values (`<script>alert(1)</script>`) do not appear in `response.error` for `POST /api/llm/feedback`, `POST /api/llm/feedback/generate-suggestion`, `POST /api/webhooks`, `PATCH /api/webhooks/:id`. 7 assertions.
- `backend/src/routes/security-regression-headers.test.ts` (extended) — asserts every nginx `proxy_set_header Host` directive uses a fixed value (no `$host` / `$http_host`), and `request-tracing.ts` source contains no `request.hostname` / `request.headers.host` reads. 2 assertions.

## Verification

| Gate | Result |
|---|---|
| `semgrep scan` (same 6 rulesets) | **0 findings** (was 12) |
| `npm run lint` | clean |
| `npm run typecheck` | clean |
| `npm run test -w backend` | 374 pass / 1 skip |
| `npm run test -w @dashboard/ai -w @dashboard/operations -w @dashboard/foundation -w @dashboard/core` | 584 pass / 39 skip (3 PostgreSQL auth failures are local-env only — `app_user` not configured locally, CI uses the real test DB) |
| New regression tests | 24 / 24 pass |

## Test plan

- [ ] CI green on `feature/* → dev` workflow
- [ ] Verify that the two new regression tests still FAIL when individually reverted (sanity check the guards aren't no-ops)
- [ ] Smoke test login + OIDC flow with the pinned `Host backend` — confirm no production code path relies on the public hostname coming from the request (it doesn't; `DASHBOARD_EXTERNAL_URL` is the supported config)
- [ ] Trace viewer still shows correct request paths after `request-tracing.ts` change (span name / `url.path` retained; only the synthetic `url.full` and `server_address` removed)

### DIFF
--- a/backend/src/routes/security-regression-error-reflection.test.ts
+++ b/backend/src/routes/security-regression-error-reflection.test.ts
@@ -1,0 +1,188 @@
+/**
+ * Security Regression — Reflected user input in error responses
+ *
+ * Validation failures must NOT echo the rejected user-supplied value back
+ * into the `error` field of the JSON response. Even though Fastify serves
+ * `application/json` (mitigating direct XSS), reflecting raw input creates
+ * a defense-in-depth gap: any client that later renders `error` via
+ * `innerHTML` / `dangerouslySetInnerHTML` would re-introduce reflected XSS,
+ * and the response becomes a content-sniffing oracle.
+ *
+ * Guards routes flagged by Semgrep rule
+ * `javascript.express.web.tainted-direct-response-express` in:
+ *   • packages/ai-intelligence/src/routes/llm-feedback.ts:101,240
+ *   • packages/foundation/src/routes/settings.ts:313,336
+ *   • packages/operations/src/routes/webhooks.ts:80,132
+ *
+ * @see https://github.com/kenhaesler/ai-portainer-dashboard/issues/1227
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import Fastify, { FastifyInstance } from 'fastify';
+import { validatorCompiler } from 'fastify-type-provider-zod';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+
+const XSS_VALUE = '<script>alert(1)</script>';
+
+// =====================================================================
+//  Static source-level guard
+// =====================================================================
+describe('Reflected input — source-level guard', () => {
+  const ROUTE_FILES = [
+    'packages/ai-intelligence/src/routes/llm.ts',
+    'packages/ai-intelligence/src/routes/llm-feedback.ts',
+    'packages/foundation/src/routes/settings.ts',
+    'packages/operations/src/routes/webhooks.ts',
+  ];
+
+  for (const rel of ROUTE_FILES) {
+    it(`${rel} does not reflect user-controlled "feature"/"event" inputs into "error" strings`, () => {
+      const abs = path.resolve(process.cwd(), '..', rel);
+      const content = readFileSync(abs, 'utf8');
+
+      // Specifically guard the patterns Semgrep flagged. We don't ban ALL
+      // template-literal `error: ` strings (some interpolate trusted
+      // constants like rate-limit thresholds); we ban the ones that echo a
+      // user-supplied feature/event back unchanged.
+      const bannedPrefixes = [
+        /error:\s*`Invalid feature:\s*\$\{/g,
+        /error:\s*`Unknown feature:\s*\$\{/g,
+        /error:\s*`Invalid event type:\s*\$\{/g,
+      ];
+      const hits = bannedPrefixes.flatMap((re) => content.match(re) ?? []);
+      expect(
+        hits,
+        `Found reflected user input pattern in ${rel}. Use a static error string + a separate \`code\` field.`,
+      ).toEqual([]);
+    });
+  }
+});
+
+// =====================================================================
+//  Runtime guard — llm-feedback routes
+// =====================================================================
+vi.mock('@dashboard/core/services/audit-logger.js', () => ({
+  writeAuditLog: vi.fn(),
+}));
+
+describe('Reflected input — llm-feedback routes', () => {
+  let app: FastifyInstance;
+
+  beforeEach(async () => {
+    const { llmFeedbackRoutes } = await import('@dashboard/ai');
+
+    app = Fastify();
+    app.setValidatorCompiler(validatorCompiler);
+    app.decorate('authenticate', async () => undefined);
+    app.decorate('requireRole', () => async () => undefined);
+    app.addHook('onRequest', async (request) => {
+      request.user = {
+        sub: 'user-1',
+        username: 'admin',
+        role: 'admin',
+        sessionId: 'sess-1',
+      };
+    });
+    await app.register(llmFeedbackRoutes);
+    await app.ready();
+  });
+
+  afterEach(async () => {
+    await app.close();
+  });
+
+  it('POST /api/llm/feedback rejects invalid feature without echoing the value', async () => {
+    const res = await app.inject({
+      method: 'POST',
+      url: '/api/llm/feedback',
+      payload: {
+        feature: XSS_VALUE,
+        rating: 'positive',
+        comment: '',
+      },
+    });
+
+    expect(res.statusCode).toBe(400);
+    const body = res.json();
+    expect(body.error).toBeTypeOf('string');
+    expect(body.error).not.toContain('<script>');
+    expect(body.error).not.toContain(XSS_VALUE);
+  });
+
+  it('POST /api/llm/feedback/generate-suggestion rejects invalid feature without echoing the value', async () => {
+    const res = await app.inject({
+      method: 'POST',
+      url: '/api/llm/feedback/generate-suggestion',
+      payload: { feature: XSS_VALUE },
+    });
+
+    expect(res.statusCode).toBe(400);
+    const body = res.json();
+    expect(body.error).toBeTypeOf('string');
+    expect(body.error).not.toContain('<script>');
+    expect(body.error).not.toContain(XSS_VALUE);
+  });
+});
+
+// =====================================================================
+//  Runtime guard — webhook routes
+// =====================================================================
+describe('Reflected input — webhook routes', () => {
+  let app: FastifyInstance;
+
+  beforeEach(async () => {
+    const { webhookRoutes } = await import('@dashboard/operations');
+
+    app = Fastify();
+    app.setValidatorCompiler(validatorCompiler);
+    app.decorate('authenticate', async () => undefined);
+    app.decorate('requireRole', () => async () => undefined);
+    app.decorateRequest('user', undefined);
+    app.addHook('preHandler', async (request) => {
+      request.user = {
+        sub: 'user-1',
+        username: 'admin',
+        role: 'admin',
+        sessionId: 'sess-1',
+      };
+    });
+    await app.register(webhookRoutes);
+    await app.ready();
+  });
+
+  afterEach(async () => {
+    await app.close();
+  });
+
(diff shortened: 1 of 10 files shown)


laboef1900:
Addressed the three info-level suggestions from the review in `a75b80b8`:

1. **Closed the residual Host-header gap in `ebpf-coverage.ts`** — `resolveOtlpEndpoint` now refuses to derive the OTLP URL from request headers when `NODE_ENV=production` and `DASHBOARD_EXTERNAL_URL` isn't set, with an explicit message. The dev fallback (used by `npm run dev` without nginx) is preserved. New regression test asserts the attacker-controlled host never reaches `deployBeyla` in production mode.
2. **Dropped `url_scheme` / `network_protocol_name` from trace spans** — `request.protocol` reads `X-Forwarded-Proto` when `trustProxy` is on, same class of issue as Host. With `server.address` already gone, the scheme alone is meaningless and just invites trace-store poisoning.
3. **Extracted `findInvalidEventType` helper in `webhooks.ts`** — POST and PATCH now call one helper instead of duplicating the validation loop.

Gates rerun: lint clean, typecheck clean, backend 374 / 1 skip, package suites pass, frontend 1828 / 1828, pre-push hook green. Static + runtime guard tests on the new code path.

PR #1230, opened 2026-05-14
fix(portainer): treat Docker 304 from start/stop as success, not failure

laboef1900:
## Summary

Docker's `/containers/{id}/start` and `/stop` endpoints are idempotent and return **HTTP 304 Not Modified** when the container is already in the requested state. `portainerFetchInner` treats every non-2xx response as an error, so `startContainer` threw `PortainerError(status=304)` in a perfectly normal case — for example when Docker auto-starts a container immediately after create, or `stopContainer` is called on an already-stopped container.

In the Beyla eBPF deploy flow this was destructive: the just-created container was actually running, but `startContainer`'s phantom 304 failure triggered the orphan-rollback path (#1228) which removed the working container. Every retry killed the previously-running Beyla.

Found by exercising the deploy against a real Edge Agent — observed the `removing just-created container to avoid orphan` log line firing immediately after a successful create, with `err: "HTTP 304: Not Modified"` from Docker.

## Change

`startContainer` and `stopContainer` now catch the 304 case via a small helper `isAlreadyInTargetStateError` and return normally — the function contract was "ensure the container is in this state," and a 304 means it already is.

Other Docker idempotent endpoints (`/pause`, `/unpause`, `/restart`, `/kill`) return 204 on success, not 304, so they're unaffected.

## Test plan

- [x] Existing 64 core tests pass.
- [x] All 76 security tests pass (deploy flow integration).
- [x] Verified end-to-end against a real Edge Agent (Type 4) — Beyla `/beyla-7` deploys cleanly to `srv-edge-01` and starts emitting OTLP traces to the configured endpoint.

## Note for reviewer

This was originally bundled into #1228 but is logically a separate Docker-API quirk fix, so it's been split into its own PR.

### DIFF
--- a/packages/core/src/portainer/portainer-client.test.ts
+++ b/packages/core/src/portainer/portainer-client.test.ts
@@ -1,6 +1,37 @@
-import { afterEach, describe, expect, it, vi } from 'vitest';
-import { decodeDockerLogPayload, sanitizeContainerLabels, _resetClientState, getCircuitBreakerStats, buildApiUrl, buildApiHeaders, pruneStaleBreakers, startBreakerPruning, stopBreakerPruning, isEdgeTunnelNotActive, EDGE_TUNNEL_NOT_ACTIVE_TOKEN } from './portainer-client.js';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Kept: undici mock — external dependency. Lets us assert behavior of
+// startContainer / stopContainer against fabricated Docker responses.
+vi.mock('undici', () => ({
+  Agent: vi.fn(function () { return { close: vi.fn().mockResolvedValue(undefined) }; }),
+  fetch: vi.fn(),
+}));
+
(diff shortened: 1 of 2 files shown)


laboef1900:
Addressed the test-coverage finding in `ee43e1ee`. Five new regression cases in `packages/core/src/portainer/portainer-client.test.ts` against a mocked `undici.fetch`:

- `startContainer` resolves on **304** (already running)
- `stopContainer` resolves on **304** (already stopped)
- `startContainer` resolves on **204** (Docker happy path — sanity check)
- `startContainer` rethrows on **404** (must NOT be swallowed by the 304 guard)
- `stopContainer` rethrows on **500** (real server failure still surfaces after retries)

These guard against the exact regression the orphan-rollback bug exhibited — if anyone changes `isAlreadyInTargetStateError` or relocates the swallow, the test suite fails instead of production losing Beyla on every retry.

Lint, typecheck clean; `@dashboard/core` portainer-client tests now 20/20 (was 15).

PR #1253, opened 2026-05-16
feat(ui): empty-state + skeleton primitives (PR 1 of 3)

laboef1900:
## Summary

Builds the foundation for the empty/loading/error-state consistency sweep:

- **Rebuilds `<EmptyState>`** with the hybrid canonical-card chrome from the design spec. Variants: `empty` / `error` / `not-configured`. No action slot (purely informational; retry / settings actions belong in the parent pane's header). The previous `EmptyState` export had 0 callers, so the rewrite is a clean break.
- **Adds five skeleton primitives** at `frontend/src/shared/components/feedback/skeleton.tsx`: `SkeletonText`, `SkeletonKpi`, `SkeletonTableRow`, `SkeletonChart`, `SkeletonList`. Each matches the shape of the eventual content. None wraps itself in card chrome — callers do that, so there's no shape jump when data arrives.
- **Demo migration on the webhooks page** (`frontend/src/features/core/pages/webhooks.tsx`) exercising both `<EmptyState>` and `<SkeletonText>` in a real page.
- **CLAUDE.md** now points future agents at the new primitives and the design spec.

## Out of scope (deliberate)

- The ~17 ad-hoc empty-state and ~25 ad-hoc loading-state call sites scattered across the app. They migrate in **PR 2** (high-traffic pages) and **PR 3** (long-tail + legacy cleanup).
- The existing `LoadingSkeleton` and `SkeletonCard` in `loading-skeleton.tsx` are untouched here — 73 active callers. They retire in PR 3.
- One **pre-existing UX bug** surfaced during review on `webhooks.tsx`: when a user filters webhooks to "Enabled" or "Disabled" with no matches, the empty state currently reads "No webhooks configured yet — Create a webhook…", which is misleading. The condition (`filteredWebhooks.length === 0`) is the same as before; this PR only swaps the visual. Distinguishing data-empty from filter-empty stays as a follow-up.

## Design spec & plan

- Spec: `docs/superpowers/specs/2026-05-16-empty-loading-states-design.md`
- Plan: `docs/superpowers/plans/2026-05-16-empty-loading-states-pr1-primitives.md`

## Test plan

- [x] `npx vitest run src/shared/components/feedback/` — 32 unit tests pass (9 EmptyState + 23 skeleton primitives)
- [x] `npx vitest run src/features/core/pages/webhooks.test.tsx` — 3/3 pass
- [x] Full frontend suite — 1911/1911 pass
- [x] `npx tsc --noEmit` — clean
- [x] `npm run lint` — clean
- [ ] Visual smoke on `/webhooks` — please verify the empty state renders with iconchip + canonical chrome, and the loading state shows 3 text bars; no layout shift when data arrives.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,4 @@ frontend/vite.config.d.ts
 # Screenshots dropped at the repo root — developer scratch, not assets.
 # `/` anchors to repo root only, leaving frontend/public/*.png tracked.
 /*.png
+.superpowers/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,6 +55,8 @@ Premium glassmorphic dashboard: bento grids, backdrop blur cards, staggered anim
 
 **Status colors:** Green=healthy, Yellow=warning, Orange=critical, Red=error, Blue=info, Gray=inactive, Purple=AI insight.
 
+**Empty / loading / error states:** Use `<EmptyState>` (variants: `empty` / `error` / `not-configured`) and the skeleton primitives (`SkeletonText`, `SkeletonKpi`, `SkeletonTableRow`, `SkeletonChart`, `SkeletonList`) in `frontend/src/shared/components/feedback/`. Skeletons live inside the caller's pane chrome — they do not wrap themselves in cards. `EmptyState` is purely informational; render any retry / settings action in the parent pane's header. See `@docs/superpowers/specs/2026-05-16-empty-loading-states-design.md` for the full rationale.
+
 For detailed specs (animation durations, easing curves, glass override patterns, layout patterns), see `@docs/ai-instructions/ui-design-system.md`.
 
 ## Testing & Mocks
--- a/docs/superpowers/plans/2026-05-16-empty-loading-states-pr1-primitives.md
+++ b/docs/superpowers/plans/2026-05-16-empty-loading-states-pr1-primitives.md
@@ -1,0 +1,957 @@
+# Empty / Loading / Error States — PR 1: Primitives Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Build the `<EmptyState>` component (variants: empty / error / not-configured) and five skeleton primitives that match content shape, so subsequent PRs can migrate the ~17 ad-hoc empty-state and ~25 ad-hoc loading-state call sites to a single canonical set.
+
+**Architecture:** Replace the dead-export `EmptyState` at `frontend/src/shared/components/feedback/empty-state.tsx` with the new hybrid-chrome design from the spec. Add a sibling `frontend/src/shared/components/feedback/skeleton.tsx` module exporting `SkeletonText`, `SkeletonKpi`, `SkeletonTableRow`, `SkeletonChart`, `SkeletonList`. Leave the existing `LoadingSkeleton` and `SkeletonCard` in `loading-skeleton.tsx` untouched — they have 73 active call sites and get retired in PR 3. Validate end-to-end by migrating one low-traffic page (`webhooks.tsx`) as a demo, which exercises both the empty state and a loading skeleton.
+
+**Tech Stack:** React 19, TypeScript strict, Vite, Tailwind v4 (custom theme via CSS vars), Vitest + jsdom + `@testing-library/react`, `lucide-react` for icons, existing `cn` util at `@/shared/lib/utils`, existing `SpotlightCard` at `@/shared/components/data-display/spotlight-card`.
+
+**Spec:** `docs/superpowers/specs/2026-05-16-empty-loading-states-design.md`
+
+**Scope:** PR 1 only. Migrations of other call sites (high-traffic in PR 2, long-tail in PR 3) are out of scope here.
+
+---
+
+## File Structure
+
+| File | Action | Responsibility |
+|---|---|---|
+| `frontend/src/shared/components/feedback/empty-state.tsx` | **Replace** | Single `<EmptyState>` component, three variants (empty / error / not-configured), no action slot, canonical card chrome with iconchip. |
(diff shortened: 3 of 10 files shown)


laboef1900:
## Code Review

### Overview

PR 1 of 3 in the empty/loading/error-state consistency sweep. Lands the shared primitives that the next two PRs will migrate ad-hoc call sites onto.

**Net change:** 9 files, +1580/-51. Two new components in `shared/components/feedback/`, one demo migration, design spec + plan docs, and a one-line CLAUDE.md pointer.

**CI:** 13/13 checks green, E2E skipped (path-based, not a regression). Mergeable.

### Strengths

- **Tightly scoped.** Resists the pull to migrate more pages while primitives are still in motion. PR boundaries match the design spec and the PR description is explicit about what's deferred.
- **Self-contained primitives.** `EmptyState` is a leaf component with a tiny prop surface (`variant`, `icon`, `title`, optional `description`/`className`) and no action slot — the design decision to push retry/settings UI to the parent pane keeps each empty state composable without a growing prop matrix. Skeleton primitives deliberately do NOT include card chrome; the caller wraps. This is the right factoring — guarantees no shape jump when data arrives and avoids double-bordering.
- **Accessibility done correctly.** Title is `<h3>` (announces as a heading). Every skeleton primitive carries `role="status" aria-label="Loading"` on the outer wrapper with a sibling `<span class="sr-only">Loading…</span>` for redundant text. `SkeletonTableRow` deliberately omits `role="status"` since it lives inside a `<tbody>` — repeating "Loading" per row would be noise.
- **Test coverage proportional to risk.** 32 unit tests for two leaf components, exercising default + variant + className passthrough + status role. No over-testing of internal DOM structure (`data-testid="empty-state-card"` rather than `.spotlight-card > div` selectors).

### Findings

#### Info

- **`webhooks.tsx` filter-vs-data empty-state copy** — `filteredWebhooks.length === 0` shows "No webhooks configured yet — Create a webhook…" whether the dataset is truly empty or just filtered to zero. This is a **pre-existing bug** (same condition, same copy intent as the prior dashed-border block); PR 1 only swaps the visual chrome. Worth fixing in PR 2 or as a follow-up — branch on `webhooksQuery.data?.length === 0` for the "configure first" copy, and inline a lighter "No matches for this filter" for the filter-empty case. Noted in the PR description.
- **Skeleton sr-only text "Loading…" (U+2026 ellipsis) differs from legacy `loading-skeleton.tsx`'s "Loading..." (three ASCII periods).** The mismatch is intentional within PR 1's scope — the legacy file has 73 callers and retires in PR 3. Worth normalizing during PR 3.
- **`SkeletonList` row markup includes a circular avatar** (`h-8 w-8 rounded-full`) which is the right shape for log/user lists but visually misleading for content without leading icons. The webhooks page migration switched to `SkeletonText` for exactly this reason; PR 2/3 reviewers should pick the primitive that matches the eventual row shape, not default to `SkeletonList` everywhere.

#### Minor

- **`bg-muted/40` vs `bg-muted/50` divergence.** New skeletons use `bg-muted/40` (slightly softer); the legacy `LoadingSkeleton` uses `bg-muted` with a shimmer overlay. Both will coexist until PR 3 retires the legacy. Document the chosen value in the spec or pick one in PR 3.
- **`EmptyState` description has a `max-w-sm` constraint** (line 42) which is sensible for centered states but quietly limits how wide the text can grow on large screens. Acceptable default; surface as a flag for any unusually long descriptive copy callers add in PR 2.

### Code Correctness, Security, Performance

- **No security implications.** Pure presentational components, no user data flows changed, no API touched. Strip-sensitive-metadata rules N/A.
- **No performance concerns.** `Array.from({ length: lines }).map(...)` runs at most ~10 iterations in any real call. Static class strings avoid Tailwind purge issues. No effects, no event handlers, no refs.
- **Type-safe.** `EmptyStateVariant` is an exhaustive union; `iconTintByVariant` is `Record<EmptyStateVariant, string>` so adding a variant without updating the map is a compile error.

### Project conventions

- ✓ Tests live alongside components (`*.test.tsx`)
- ✓ Path alias `@/shared/...` used consistently
- ✓ `cn` from `@/shared/lib/utils` for class merging
- ✓ Lucide icon name collision resolved via `Webhook as WebhookIcon` (matches `Users as UsersIcon` pattern elsewhere)
- ✓ No `--no-verify` flags, no `.env` or credentials touched
- ✓ Branch model respected (`feature/ui-empty-loading-primitives → dev`)
- ✓ CLAUDE.md updated with a pointer to the new primitives + spec

### Recommendation

Approved with informational findings. The Info items are deliberate scope deferrals documented in the PR description, not defects.

laboef1900:
## Re-Review (commit `d5c644b2`)

### What changed

- `webhooks.tsx`: split the empty branch. `<EmptyState>` now renders only when the *dataset* is empty (`webhooksQuery.data?.length === 0`). When data exists but the filter matches nothing, an inline `<p>` caption ("No webhooks match the current filter.") replaces it — communicates that the absence is filter-driven, not configuration-driven.
- `webhooks.test.tsx`: refactored the static `useWebhooks` mock into a hoisted `vi.fn()` so per-test overrides work. Added a regression test that mocks `data: []` and asserts on the EmptyState `data-testid` plus the absence of the filter caption.

### Verification

- Local: 1912 frontend tests pass (was 1911 — one new test added). `tsc --noEmit` clean. `npm run lint` clean.
- A filter-empty *interaction* test was considered but dropped — driving the Radix-based `ThemedSelect` from a unit test is brittle and high-cost for one assertion. The branch's correctness is self-evident from a six-line conditional, and the test that does exist guards the more important regression (truly-empty data path).

### Findings status

- **#1 Filter-vs-data empty distinction** — fixed (this commit).
- **#2 `Loading…` ellipsis mismatch with legacy `loading-skeleton.tsx`** — out of PR 1 scope. Legacy file is touched in PR 3 with its 73 callers; normalize there.
- **#3 SkeletonList avatar shape** — not a defect; was a usage note for PR 2/3 reviewers. Already applied (this PR uses SkeletonText for webhooks).
- **#4 `bg-muted/40` vs `bg-muted/50` divergence** — out of PR 1 scope. PR 3 concern.
- **#5 `max-w-sm` description width** — marked acceptable in the original review.

### Recommendation

Approved. All in-scope findings are resolved; deferred items are properly demarcated for PR 3.

Issue #1298, opened 2026-05-27
Feature: Health & Monitoring — "Mark as false positive" feedback loop


laboef1900:
Child of epic #1291.

## Problem Statement

The system has no way to learn from operator feedback today. When an operator sees a clearly-false anomaly, they can dismiss it locally but the dismissal isn't recorded, isn't attributed, and doesn't inform future tuning. The feedback gap also blocks evidence-based decisions about which detectors are misfiring.

Corresponds to fix 10 / PR E in epic #1291.

## Proposed Solution

Add a "Mark as false positive" action on each `CorrelatedAnomalyCard`. Record the feedback in a new `anomaly_feedback` table. Surface the **per-detector false-positive rate** as a small badge in the Settings panel (or in the existing Health & Monitoring header), so operators can see whether their feedback is moving the needle.

### Files touched

| File | Change |
|---|---|
| `frontend/src/features/ai-intelligence/pages/ai-monitor.tsx` (`CorrelatedAnomalyCard`) | Add a "Mark as false positive" button next to the existing action affordances. Optimistic update on click. |
| `packages/ai-intelligence/src/routes/monitoring.ts` | New `POST /api/monitoring/anomaly-feedback` (records a feedback row); new `GET /api/monitoring/anomaly-feedback/rates` (returns per-detector false-positive rate). Both `fastify.authenticate`-gated. |
| DB migration | New `anomaly_feedback` table — see schema below. |

### `anomaly_feedback` table

| Column | Type | Notes |
|---|---|---|
| `id` | `serial` PK | |
| `anomaly_id` | foreign key into anomalies | indexed |
| `user_id` | foreign key into users | indexed |
| `disposition` | text enum (`false-positive` for now; future values reserved: `true-positive`, `unsure`) | |
| `created_at` | timestamptz | default `now()` |
| | | UNIQUE `(anomaly_id, user_id)` — see "Multi-user feedback semantics" below |

### Multi-user feedback semantics

A real design question: should multiple users be able to file feedback on the **same** anomaly?

**Recommendation:** yes — one feedback row per `(anomaly_id, user_id)` pair (enforced by the unique index). Resubmitting your own feedback is idempotent (the same row updated, or a no-op insert with `ON CONFLICT DO NOTHING` and the existing `created_at` preserved). Multiple users may legitimately disagree about whether an anomaly was real, and the rate calculation needs the per-user signal anyway.

Document this in the migration comment so it doesn't drift.

[...]

Issue #1291, opened 2026-05-27
Epic: Reduce ML-Detected Anomaly false positives on Health & Monitoring


laboef1900:
> **Epic.** This issue tracks the cross-component work to reduce false-positive rates on the Health & Monitoring page. The original spec listed 10 fixes; per the repo's "one concern per issue" guidance and the project's `epic` label convention, the work has been split into 5 children listed below. Each child is independently shippable. Close this epic when all five are closed.

## Children

| Order | Issue | Scope | Status |
|---|---|---|---|
| 1 | #1294 | **Runtime audit + config-default adjustments** (fixes 1, 4, 5, 7, 8) — fastest noise reduction; the audit informs the ordering of the rest. | open |
| 2 | #1295 | **CV variance scaling + hour-of-day baseline** (fixes 2 + 3) — largest single algorithmic win. | open |
| 3 | #1296 | **Correlated suppression** (fix 6) — collapse co-occurring p95 + error-rate spikes into one record. | open |
| 4 | #1297 | **Sensitivity preset UI + per-user persistence** (fix 9) — Low / Default / High control in Health & Monitoring. | open |
| 5 | #1298 | **False-positive feedback loop + table** (fix 10) — "Mark as false positive" action + per-detector rate badge. | open |

Suggested order matches the order above. The runtime audit in #1294 may surface a reason to swap #1295's internal trace-vs-metric ordering, but the epic-level sequence (children 1 → 5) should not change without a comment here explaining why.

## Problem Statement

The dashboard's Health & Monitoring page surfaces ML-Detected Anomalies that fire **too aggressively** in practice — operators report a high false-positive rate, which trains them to ignore the panel and erodes trust in the AI intelligence layer. The detector is **observer-first and read-only**, so the consequence is not bad actions taken — it's signal fatigue: real incidents get lost in the noise.

Concrete pain points observed today:

- Stable services with a 4-5% baseline error rate alternate between "anomalous" and "fine" because the trace detector applies a **z-score of 2.5** (stricter than the metric detector's 3.5) layered on top of a hard **5% error-rate floor** with no variance scaling.
- Containers with daily traffic cycles (e.g. nightly batch jobs, morning ramp) trip anomalies during the ramp because the metric detector only looks at a **20-sample moving window** with no seasonality / hour-of-day baseline.
[...]

Issue #1236, opened 2026-05-14
Feature: Latency and error-rate anomaly detection driven by trace data


laboef1900:
## Problem Statement

`packages/ai-intelligence/src/services/monitoring-service.ts` only consumes container-level metrics (CPU, memory, network IO). It cannot detect a slow downstream database or a spike in 5xx responses because that signal never reaches it. Beyla now provides authoritative per-route latency and status — feeding it to the detector closes the most common production failure mode (latency before resource saturation).

## Proposed Solution

Extend the existing adaptive-anomaly + isolation-forest pipeline with two new metric families sourced from spans:

| Series | Aggregation | Trigger |
|---|---|---|
| `span.latency.p95` | per (service, route, 1m) | adaptive z-score on rolling 24h baseline |
| `span.error_rate` | per (service, 1m) | adaptive z-score + absolute threshold (>5%) |

New anomalies emitted with `metricType: 'latency_p95'` or `metricType: 'error_rate'` so the existing UI/predictive-alert plumbing picks them up unchanged.

## Use Cases

1. **Postgres slowed down** — CPU is fine, but every API call's p95 jumped from 20ms to 800ms. Detector fires within 2 minutes.
2. **Bad deploy** — Error rate goes from 0.1% to 4% on one route. Detector fires; the alert links to the affected trace.
3. **Cross-service correlation** — Anomalies tagged with the source container's `container.id` so the existing dashboard can render them alongside metric anomalies on the same timeline.

## Acceptance Criteria

- [ ] New `packages/ai-intelligence/src/services/trace-anomaly.ts` module pulling from the RED endpoint (#TBD) or directly from `spans`.
- [ ] Monitoring service wires the new series into its existing cycle (`monitoring-service.ts:Starting monitoring cycle`).
- [ ] Anomalies populate the same predictive-alerts table; UI requires no change.
- [ ] Configurable thresholds via env: `TRACE_ANOMALY_P95_ZSCORE` (default 2.5), `TRACE_ANOMALY_ERROR_RATE_PCT` (default 5).
- [ ] Test: seed spans → run detector → assert anomaly written with correct attributes.

## Technical Considerations

- Architecture: reuses `adaptive-anomaly` and `isolation-forest-detector` modules already shipped.
- Dependencies: requires RED endpoint (#TBD) or equivalent SQL query.
- Storage: writes to the existing anomaly table, no schema change.
[...]

PR #1303, opened 2026-05-27
feat(anomaly): raise z-score / window / contamination defaults + audit + rate limit

### LINKED ISSUE #1294: Feature: Anomaly detection — runtime audit + config-default adjustments
Child of epic #1291.

## Problem Statement

Five of the ten anomaly-tuning fixes in epic #1291 are mechanical config / small-code changes that should ship first to give operators the fastest noise reduction. This issue bundles them with the runtime audit that orders the subsequent algorithmic children.

In scope (corresponds to PR A in #1291):

- **Runtime audit** — identify which detector currently dominates false-positive volume so the algorithmic children (#1291-B/C/D/E) ship in the order that reduces noise fastest.
- **Fix 1** — raise `TRACES_ANOMALY_P95_ZSCORE` default from `2.5` → `3.0`.
- **Fix 4** — raise `ANOMALY_MOVING_AVERAGE_WINDOW` default from `20` → `60`.
- **Fix 5** — lower `ISOLATION_FOREST_CONTAMINATION` default from `0.15` → `0.05`.
- **Fix 7** — per-service rate limit (max 1 new anomaly per service per 5 min) layered on top of the existing 10-min per-key cooldown.
- **Fix 8** — minimum baseline sample count on the trace detector, mirroring `ANOMALY_MIN_SAMPLES`.

### Runtime audit (do this first)

Commit a SQL query or short Node script (`scripts/audit-anomaly-detectors.ts` or a documented psql snippet) that:

1. Groups anomalies from the last 7-14 days by detector source and severity.
2. Estimates the false-positive share per detector using whichever proxy actually exists in the schema:
   - Anomalies dismissed / acknowledged with a `false-positive` disposition, if a disposition column already exists.
   - Anomalies acknowledged but with no follow-up remediation action recorded.
   - Anomalies that resolved themselves within the cooldown window without any human action.
3. Outputs a results table:

   | Detector | Total anomalies (7d) | Estimated false-positive share | Suggested ordering for #1291-B…E |
   |---|---|---|---|

Commit the results table to a design note in `docs/superpowers/specs/` or `docs/ai-instructions/`. Use it to comment the suggested ordering back on the parent epic #1291 — the audit informs ordering of the other children, not their scope.

If the data is inconclusive (all three detectors roughly equal), default to the order documented in #1291.

### Code-level fixes

| # | Change | Files |
|---|---|---|
| 1 | `TRACES_ANOMALY_P95_ZSCORE` default `2.5` → `3.0` | `packages/core/src/config/env.schema.ts`, `.env.example`, `CLAUDE.md` |
| 4 | `ANOMALY_MOVING_AVERAGE_WINDOW` default `20` → `60` | `packages/core/src/config/env.schema.ts:85`, `.env.example`, `CLAUDE.md` |
| 5 | `ISOLATION_FOREST_CONTAMINATION` default `0.15` → `0.05` | `packages/core/src/config/env.schema.ts`, `.env.example`, `CLAUDE.md` |
| 7 | Per-service rate limit on top of per-key cooldown | `packages/ai-intelligence/src/services/trace-anomaly.ts` (cooldown logic near line 62) |
| 8 | Minimum baseline sample count enforced on trace path | `packages/ai-intelligence/src/services/trace-anomaly.ts` |

## Acceptance Criteria

- [ ] Runtime audit script / query checked in, with a results table committed in `docs/superpowers/specs/` or `docs/ai-instructions/`. A comment on parent epic #1291 records the resulting recommended ordering for children B-E.
- [ ] **Fix 1:** `TRACES_ANOMALY_P95_ZSCORE` default is `3.0` in `env.schema.ts`, `.env.example`, and `CLAUDE.md`.
- [ ] **Fix 4:** `ANOMALY_MOVING_AVERAGE_WINDOW` default is `60` in the same three files.
- [ ] **Fix 5:** `ISOLATION_FOREST_CONTAMINATION` default is `0.05` in the same three files. A regression test in `packages/ai-intelligence/src/services/isolation-forest-detector.test.ts` covers a realistic mixed-noise scenario: train on a **deterministic** seeded series of ~95% quiet samples (Gaussian noise σ=0.1 around a constant mean) plus ~5% outliers (the outliers seat the threshold above the noise floor, matching real production where bona fide anomalies exist), then evaluate on a fresh 1000-sample quiet-only series and assert ≤ 1% false-positive rate. The seed and series shape are documented in the test. _Note: an earlier version of this AC asked for ≤ 1% on a pure-noise series, but `IsolationForest.fit()` calibrates the threshold to the configured contamination percentile by construction — so on pure noise the FP rate floors at ~contamination. The mixed-noise reframe is the property we actually want to guard._
- [ ] **Fix 7:** Per-service rate limit (default: max 1 new anomaly per service per 5 min) layered on top of the existing 10-min per-key cooldown, with a burst-case test that fires 10 anomalies for the same service within 1 minute and asserts only the first is persisted.
- [ ] **Fix 8:** Minimum baseline sample count enforced on the trace detector (mirroring `ANOMALY_MIN_SAMPLES`), with a warm-up test that simulates a brand-new service receiving < `ANOMALY_MIN_SAMPLES` data points and asserts no anomalies fire.
- [ ] All existing tests in `packages/ai-intelligence/` continue to pass.
- [ ] Docs updated: `docs/architecture.md`, `.env.example`, `CLAUDE.md`. Design note (with audit results) shipped.

## Technical Considerations

- **Determinism in tests.** Any synthetic-series test must seed its RNG explicitly (e.g. a Mulberry32 / xoroshiro128 with a fixed seed) — do not rely on `Math.random()` or `Date.now()`. Document the seed in the test so flakiness can be reproduced.
- **Observer-first:** Read-only. No mutating actions.
- **RBAC:** No new routes here.
- **Backwards compatibility:** All env var names retained; only defaults change. Operators with pinned values in `.env` are unaffected.

**Effort Estimate:** Small-Medium
**Impact Estimate:** High (fastest noise reduction in the epic)
**Priority Score:** 8.0/10

laboef1900:
Closes #1294. Part of epic #1291.

## Summary

Five fixes from epic #1291 child A, shipped together because they all reduce noise on the Health & Monitoring page via the same code paths:

- **Fix 1:** `TRACES_ANOMALY_P95_ZSCORE` default `2.5 → 3.0` — at 2.5σ Gaussian baselines yield ~1.2% false positives per sample (≈17/day for a 1-sample-per-min service); 3.0σ drops that to ~0.27% (≈4/day).
- **Fix 4:** `ANOMALY_MOVING_AVERAGE_WINDOW` default `20 → 60` — 20 samples (~20 min) over-reacts to traffic ramps; 60 samples (~1h) smooths short bursts while preserving sensitivity to sustained shifts.
- **Fix 5:** `ISOLATION_FOREST_CONTAMINATION` default `0.15 → 0.05` — contamination calibrates the IF threshold to flag the top fraction of training scores; 0.15 forces ~15% of every stable workload's points into the anomaly class by construction.
- **Fix 7:** per-service rate limit (`TRACES_ANOMALY_PER_SERVICE_MIN`, default 5 min) layered on top of the existing 10-min per-(service,metric_type) cooldown in `trace-anomaly.ts`. Caps a single flapping service to one anomaly per window regardless of metric_type.
- **Fix 8:** trace-path baseline warm-up via new `TRACES_ANOMALY_MIN_SAMPLES` (default 10, mirrors `ANOMALY_MIN_SAMPLES`). Brand-new services no longer fire on their first hour of data.

Audit committed at [`docs/superpowers/specs/2026-05-27-anomaly-detector-audit.md`](docs/superpowers/specs/2026-05-27-anomaly-detector-audit.md) with results table; the recommended ordering for #1291 children B–E confirms the epic's existing order. Live SQL/script at [`scripts/audit-anomaly-detectors.ts`](scripts/audit-anomaly-detectors.ts).

**Drive-by fix:** this PR also corrects pre-existing doc drift on `ANOMALY_ZSCORE_THRESHOLD` in `docs/configuration.md` (was `2.5`, schema default is `3.5`). Surfaced incidentally while auditing the anomaly defaults; included here so the docs match the running config.

## Test plan

- [x] Lint / typecheck clean (`npm run lint`, `npm run typecheck`)
- [x] Deterministic seeded synthetic test for fix 5 — `isolation-forest-detector.test.ts` uses Mulberry32 (seed `0x9E3779B9`), 950 quiet Gaussian samples + 50 outliers for training, 1000 fresh clean samples for evaluation. Asserts ≤ 1% anomaly rate. Pre-fix (contamination=0.15) the fixture fails by a wide margin.
- [x] Burst-case test for fix 7 — 10 trace-anomaly cycles within 1 minute for the same flapping service → only the first insight persists.
- [x] Upper-bound test for fix 7 — `vi.useFakeTimers()` advances past `TRACES_ANOMALY_PER_SERVICE_MIN + COOLDOWN_MS` and verifies the next anomalous cycle DOES fire again, locking in both bounds of the rate limit.
- [x] Warm-up test for fix 8 — service with only 5 baseline buckets (< `TRACES_ANOMALY_MIN_SAMPLES = 10`) → asserts `expect(insertSpy).not.toHaveBeenCalled()` (clearer intent than the previous indirect loop).
- [x] Audit results committed to `docs/superpowers/specs/2026-05-27-anomaly-detector-audit.md`
- [x] Existing `trace-anomaly.test.ts` cases still green after the new defaults / mock-config update
- [x] Existing `packages/core/src/config/index.test.ts` defaults assertions updated to the new values
- [x] All non-DB tests in `packages/ai-intelligence` and `packages/core` pass. DB-backed tests (`incidents-*.test.ts`, `insights-store.test.ts`, etc.) fail with a pre-existing `password authentication failed for user "app_user"` against the local PostgreSQL on port 5433 — reproducible on `dev` without these changes, so not caused by this PR. The pre-push hook against the contributor's CI-equivalent harness ran 1956/1956 tests passing.

## Notes for reviewers

- **`is_acknowledged`-based FP proxy in the audit.** The current schema has no explicit `false_positive` disposition column (that lands in #1298). Until then the audit estimates FP share using `is_acknowledged` + `incidents.self-resolved-within-10min` as proxies. Numbers are directional, not measurements — caveat documented in the design note.
- **Synthetic data caveat.** Dev has produced zero trace anomalies since #1236 merged, so the audit numbers in the design note are a synthetic projection. Re-run `scripts/audit-anomaly-detectors.ts` against prod after ≥ 7 days of post-rollout data and revise the table.
- **Backwards compatibility.** All env var names are retained; only defaults change. Operators with pinned values in their `.env` are unaffected.


laboef1900:
## Review — PR #1303 (feat(anomaly): defaults + audit + rate limit)

### Overview
This PR is a tight, well-justified noise-reduction pass: five default/behaviour changes on the anomaly detectors plus a runtime audit script and an excellent design note. The implementation is small, the test coverage on the new behaviour is real (not smoke), and docs/env.example/CLAUDE.md/configuration.md are kept consistent.

### Strengths
- Audit script uses parameterised `:days` substitution (server-side `INTERVAL :days days` interpolated client-side as a literal integer parsed via `Number()` + `/^\d+$/` regex — safe, no string concat from user input). `--print-sql` mode is a nice operational touch; doesn't leak secrets.
- Design note is one of the better ones I've seen — explicitly calls out the synthetic-data caveat, the `is_acknowledged` FP proxy's over/undercount failure modes, and documents the Mulberry32 seed choice with a reference.
- Per-service rate limit is atomic with the existing per-key cooldown (single `markInserted(seriesKey, service)` updates both maps); `__resetTraceAnomalyLogState()` is updated to clear `lastServiceInsertAt`.
- Warm-up branch covers both latency_p95 (`>= Math.max(3, minBaselineSamples)`) and error_rate (early `continue` if `errSeries.length < minBaselineSamples`).
- DB-backed test failures the author flags as pre-existing: confirmed — PR touches none of `incidents-*.test.ts` / `insights-store.test.ts`, and the failure mode (`password authentication failed for user "app_user"` on port 5433) is a local DB config issue, not a code regression.

### Concerns / suggestions

**Fix-5 test reframing — verdict: 🔵 Info, the author is technically correct.**

`IsolationForest.fit()` literally sorts training scores descending and picks `floor(N * contamination) - 1` as the threshold (`isolation-forest.ts:120-123`). On pure same-distribution noise, fresh samples will hit the anomaly class at rate ≈ contamination by construction — **the original AC ("≤ 1% with contamination=0.05") is mathematically unsatisfiable on noise-only data**. The 950+50 training mix is the right reframe: it seats the threshold above the noise floor exactly as production data would, then evaluates on fresh clean samples. The test verifies the intended property (stable workload does not trip), not the literal AC wording. Recommend **updating issue #1294's AC text** to match the test design rather than asking for a redo. The design note already explains the rationale (lines 188-198) — that's the right place; just sync the issue.

**🔵 Info — pre-existing `ANOMALY_ZSCORE_THRESHOLD` doc drift.** `docs/configuration.md` previously listed default `2.5` but the schema default has been `3.5` for some time. The PR silently corrects this (line 74 of the diff). Worth a one-line mention in the PR body as a drive-by fix so it doesn't look accidental.

**🔵 Info — per-service rate-limit map is unbounded.** `lastServiceInsertAt` grows by one entry per distinct service name and is never pruned (same pattern as the existing `lastInsertedAt`, which is also unbounded). Not a problem at realistic fleet sizes but worth a TODO comment if the trace path ever ingests user-controlled span attributes for grouping.

**🔵 Info — burst-case test is convincing but only proves the lower bound.** It verifies "exactly 1 insight" across 10 cycles, which is the right assertion. Optional: a complementary test that *advances time past* `TRACES_ANOMALY_PER_SERVICE_MIN` (e.g. via `vi.useFakeTimers()`) and confirms a second anomaly *does* fire would lock in the upper bound. Not required for merge.

**🔵 Info — warm-up assertion could be tighter.** The test currently iterates `spy.mock.calls` asserting each is `[]`. If the gate were silently bypassed the spy might be called with a non-empty array, which the loop would catch — but `expect(spy).not.toHaveBeenCalledWith(expect.arrayContaining([expect.anything()]))` or simply `expect(spy).not.toHaveBeenCalled()` (since no other code path inserts here) would be clearer intent.

No 🟡 / 🔴 findings.

### Recommended label
`PR:Info`

PR #1322, opened 2026-05-29
fix: prevent Portainer cache poisoning from failed fetches

laboef1900:
## Problem

Background update tasks repeatedly fail with \\"Background update failed: Unable to connect to Portainer\\" despite Portainer being online. Logs show:

- `Circuit breaker opening — failure threshold reached` for `portainer-api:endpoint-81`
- `Cannot read properties of undefined (reading 'map')` at `endpoints.js:24`
- `Cannot read properties of undefined (reading 'filter')` at `security-audit.js:92`
- `Circuit breaker probe failed — re-opening circuit` (recovery also fails)

## Root cause

A two-part cascading bug:

1. **Cache poisoning by `undefined\** (core)**: When `cachedFetch` encounters a fetch failure (circuit-breaker tripping), the failed promise is cached. On the next call, `cache.get()` returns `undefined`, which callers then call `.map()` or `.filter()` on — crashing the process. The cache entry persists across container restarts if Redis has the stale entry.

2. **No null guards in callers**: `endpoints.ts` and `security-audit.ts` assumed the cached value was always an array, so any cache miss/undefined propagated as `TypeError`.

## Fix

**`packages/core/src/portainer/portainer-cache.ts`**
- `cachedFetch`: Only cache when `data !== undefined`; on fetch failure, invalidate the cache key so the next call retries
- `cachedFetchSWR`: Same guards on both L1 and L2 revalidation paths (3 locations)

**`packages/foundation/src/routes/endpoints.ts`**
- `endpoints.map()` → `(`endpoints` || []).map()` (2 locations)

**`packages/security/src/services/security-audit.ts`**
- `endpoints.filter()` → `(endpoints` || []).filter()`

## Verification

Applied to production container-insights backend. After fix + restart:
- Circuit breaker recovered successfully (`probe succeeded — closing circuit`)
- Zero `undefined` TypeError crashes
- Zero circuit breaker re-open failures
- Full monitoring cycle running (68 containers, 1 insight)

### DIFF
--- a/packages/core/src/portainer/portainer-cache.test.ts
+++ b/packages/core/src/portainer/portainer-cache.test.ts
@@ -690,6 +690,219 @@ describe('stale-while-revalidate (cachedFetchSWR)', () => {
   });
 });
 
+describe('undefined / failed-fetch poisoning guard (#1270)', () => {
+  beforeEach(() => {
+    vi.resetModules();
+  });
+
+  // ── 2a: a fetcher resolving undefined must NOT populate the cache ──────────
+
+  it('cachedFetch does not cache an undefined resolution (no poisoned entry)', async () => {
+    vi.doMock('../config/index.js', () => ({
+      getConfig: () => ({ ...baseConfig }),
+    }));
+    vi.doMock('redis', () => ({
+      createClient: vi.fn(),
+    }));
+
+    const { cachedFetch, cache } = await import('./portainer-cache.js');
+    const setSpy = vi.spyOn(cache, 'set');
+
+    const undefinedFetcher = vi.fn().mockResolvedValue(undefined);
+    const result = await cachedFetch('poison:cf', 30, undefinedFetcher);
+
+    // The undefined result is surfaced to the caller…
+    expect(result).toBeUndefined();
+    // …but never written to the cache.
+    expect(setSpy).not.toHaveBeenCalled();
+    // A direct read confirms no poisoned entry was persisted.
+    expect(await cache.get('poison:cf')).toBeUndefined();
+
+    // A second call re-invokes the fetcher (cache was not poisoned with undefined).
+    await cachedFetch('poison:cf', 30, undefinedFetcher);
+    expect(undefinedFetcher).toHaveBeenCalledTimes(2);
+  });
+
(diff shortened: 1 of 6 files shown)


laboef1900:
## Automated code review — findings addressed ✅

Reviewed the cache-poisoning fix. Findings below were resolved in `1a7e5bb4` and independently re-verified against the pushed commit.

**Findings (original PR):**
- ⚠️ **Guard missed the primary path** — in `security-audit.ts` the `(endpoints || [])` guard was only on the `endpointId`-set branch; the "audit all endpoints" branch passed `undefined` straight into `.filter()` → `TypeError`.
- ⚠️ **No tests** — the mandatory regression coverage for the poisoning scenario was missing.
- **Log regression** — the inner L2 SWR `warn` lost its "L2" qualifier (became identical to the L1 site).
- **Root-cause clarity** — the stated cause didn't match the control flow (the circuit breaker throws; thrown fetches reject without caching).
- **Nit** — `invalidate()` cleared only the plain key, not the `:gz` compressed variant.

**Resolution (`1a7e5bb4`):**
- Guards moved to the source assignment (`const endpoints = (await cachedFetch(...)) ?? []`) in `security-audit.ts` and `endpoints.ts`, covering every branch/use-site.
- 11 regression tests added (core, security, foundation): `undefined` is never cached (blocking + SWR), a good cached value survives a throwing fetch with the error surfaced, `invalidate` is called on failure, and the audit-all/endpoints routes return `[]` instead of crashing.
- "L2" qualifier restored.
- Root cause confirmed: the real `undefined` source is `portainerFetchInner` on HTTP 204/empty body; Zod-parsing fetchers throw rather than caching `undefined`, so the load-bearing protection is the pre-existing `if (data !== undefined)` write-guard (correctly retained; Zod fetchers correctly untouched).
- `invalidate()` now deletes both the plain and `:gz` keys.

**Verification:** independent re-review of `1a7e5bb4` confirmed all findings resolved. Tests: core **53/53**, security **13/13**, foundation **12/12**; root typecheck clean.

**Minor (non-blocking):** one redundant cache-hit test (`retain:cf`) doesn't exercise the failure path — but that path is already covered by a separate throwing-fetch test.

**Assessment: Approved** (`PR:Approved`).

Issue #1420, opened 2026-06-01
E2E smoke suite fails in CI (22/41): login-spec drift + tests require a populated Portainer


laboef1900:
## Summary

With the nightly E2E backend-startup blocker fixed (#1417 — weak `DASHBOARD_PASSWORD`/`REDIS_PASSWORD`/`TIMESCALE_PASSWORD` rejected under `NODE_ENV=production`), the full Playwright stack now comes up **healthy** and the E2E suite actually runs for the first time — and it fails **22 of 41 tests** (19 passed). The failures fall into two unrelated categories. Because the backend never booted in CI before #1417, this suite has effectively **never passed in CI**.

Failing run: https://github.com/kenhaesler/ai-portainer-dashboard/actions/runs/26737112237 (all containers `Up (healthy)`; 19 passed / 22 failed, 20.5m).

## Root Cause

**Category A — E2E specs drifted from the current login UI.** The smoke tests assert a login heading and field labels that no longer match the "cinematic" login page.

In `e2e/smoke.spec.ts` line 11:
```typescript
await expect(page.getByRole('heading', { name: /sign in|login/i })).toBeVisible({ timeout: 10_000 });
```
This fails *before any login or data fetch* — the heading isn't found. The other two smoke tests then fail at `getByLabel(/username/i)` (lines 25 / 40). The login page is static React served by a healthy frontend, so this is selector/copy drift, not data.

**Category B — data-dependent tests need a populated Portainer that CI doesn't provide.** CI points the backend at a dead dummy:

In `.github/workflows/ci.yml` (e2e job env):
```yaml
PORTAINER_API_URL: http://127.0.0.1:9999
PORTAINER_API_KEY: e2e-dummy-key
```
Nothing listens on :9999, so the Portainer circuit breaker opens (`CircuitBreakerOpenError`, benign for `/health`) and pages render no endpoints/containers/stacks/groups. Tests asserting that data then time out, e.g. `e2e/workload-explorer-dropdown-position.spec.ts` line 111:
```typescript
const trigger = page.locator('#group-select');
await expect(trigger).toBeVisible({ timeout: 15_000 }); // element(s) not found
```

## Steps to Reproduce

1. Trigger the `CI` workflow's E2E job (nightly schedule, manual dispatch, or a PR with the `e2e` label).
2. The stack builds and `docker compose up --wait` succeeds — all services healthy.
3. `npm run test:e2e` runs.
[...]

PR #1428, opened 2026-06-03
fix(frontend): isolate page render errors so the app shell survives a page crash

laboef1900:
## Summary

**Defense-in-depth, independent of #1420** (this does NOT fix the E2E failures — that's #1430 + #1431 + #1432; combined run 26959256777 confirmed this boundary changed nothing there).

`AppLayout` wraps the router `<Outlet>` in an `ErrorBoundary` (`PageBoundary`), placed **below** the layout and inside the route-keyed wrapper. A render error in one page now degrades to an inline error card while the **sidebar + header stay mounted**, instead of bubbling to the `/` route's `errorElement` and replacing the whole shell. The boundary resets on navigation.

**Scope (trimmed on review):** originally this PR also carried a 10-page "renders without throwing on empty/unavailable fleet" test sweep — but that sweep found **no actual bugs** (every page already guarded its data), so it was removed (~370 low-signal lines). What remains is the structural value:
- `frontend/src/features/core/components/layout/app-layout.tsx` — the `PageBoundary` wrapper (+21 lines).
- `frontend/src/features/core/components/layout/app-layout.test.tsx` — an integration test proving a throwing child route keeps `[data-testid="sidebar"]` mounted and shows the fallback.
- `docs/architecture.md` — a one-line note.

(The design/plan docs under `docs/superpowers/` are retained as the historical Category-B record; #1432 supersedes their diagnosis.)

Merge on its own merits — not required for #1420.

## Test Plan
- [x] `app-layout.test.tsx` — 1/1 (throwing child route → sidebar survives + fallback shown)
- [x] `frontend` suite green (the reverted page tests are back to their `dev` versions)


laboef1900:
Reframing: this is an independent **defense-in-depth** improvement (a page render error degrades to an inline card instead of unmounting the sidebar/header), discovered during the #1420 investigation but **not** the fix for the #1420 E2E failures. Combined-branch E2E run 26896200138 confirmed the real #1420 blockers are an nginx `/health` SPA-route collision and smoke's cached-auth test design — neither of which this PR addresses. Keeping it on its own merits (it has a passing unit test); de-linking from #1420.

PR #1459, opened 2026-06-11
security: harden against audited attack vectors (SSRF, RBAC, session revocation, prompt-guard)

laboef1900:
## Summary

Proactive whole-application security hardening. A multi-dimension security audit (auth/RBAC/IDOR, JWT/session, OIDC, SSRF, path traversal, LLM-guard coverage, input validation, secrets disclosure, WebSockets, infra/headers/CORS, crypto, machine-ingestion endpoints, frontend XSS) surfaced **20 verified, un-mitigated gaps** after each candidate was adversarially re-verified for exploitability and against existing controls (false positives / already-defended items were filtered out). Every fix ships with a regression test.

No single tracking issue — this is a proactive audit rather than a reported bug.

## Fixes

**High**
- **LLM token exfiltration / SSRF** — `/api/llm/test-connection` and `/api/llm/models`: admin-gated; the stored provider token is never forwarded to a caller-supplied host; the `host` override is admin-only. (Internal/self-hosted LLM testing still works — private IPs are intentionally *not* blanket-blocked there; the boundary is the admin gate + no-token-forwarding.)
- **Stale-privilege JWTs** — role change, password reset, user deletion, and OIDC group-mapping downgrades now revoke the affected user's sessions immediately (the role lived only in the signed JWT and was never re-checked).

**Medium**
- SSRF-guard bypasses fixed (`0.0.0.0/8`, bracketed IPv6 `[::1]`, IPv4-mapped IPv6 `[::ffff:169.254.169.254]`); webhooks re-validate the destination at delivery time.
- WebSocket chat: tool-result content and client `context` are guarded before reaching the model (second-order prompt injection); malformed payloads no longer cause an unhandled rejection.
- `/api/incidents` pagination is now bounded (unbounded-`limit` DoS).
- Live sockets are periodically re-validated — revocation/demotion disconnects them instead of surviving until JWT expiry.
- Socket.IO JWT is kept out of nginx access logs.

**Low / defense-in-depth**
- Admin-gate `/health/ready/detail`; operator-gate insight acknowledge; `no-store` on `/api/users` and `/api/backup`.
- Constant-time token comparison (Prometheus, trace ingest) + production trace-ingest key validation; OTLP recursion depth cap.
- Username-enumeration timing fix; OIDC username-collision no longer leaks raw DB errors; llm-feedback comments pass the prompt guard.
- Global error handler stops raw 5xx error reflection in production.

## Known residual
A holder of the shared trace-ingestion API key can still forge service identities in spans (write-only impact). The ingestion boundary is hardened (constant-time compare + production key-strength validation) and documented as a write-trust boundary in `docker/.env.example`; a global cross-identity span cap would be a larger feature change.

## Operational note
With `TRACES_INGESTION_ENABLED=true`, production now **refuses to start** if `TRACES_INGESTION_API_KEY` is shorter than 16 chars (mirrors `PROMETHEUS_BEARER_TOKEN`). Documented in `docker/.env.example`.

## Testing
core **815**, ai-intelligence **1032**, observability **253**, operations **188**, foundation **45**, backend security-regression **216** — all passing. Build, typecheck, and lint clean.


laboef1900:
## Review findings addressed (commit `1536250f`)

**1. Scope / commit bundling** — PR base retargeted from `dev` to `feature/security-llm-guard-chokepoint` (PR #1458), which already owns the `chatStream` commit. This PR now shows **only the 35 hardening files (+1142)**. Once #1458 merges to `dev`, this can be retargeted/rebased onto `dev`.

**2. `users.ts` revocation robustness** — `invalidateAllUserSessions` on role/password/delete is now wrapped in a best-effort `try/catch` with a warn log (matching the `oidc.ts` pattern), so a transient revocation failure no longer turns an already-committed update into a 500.

**3. Four missing regression tests added:**
- **Webhook delivery-time SSRF re-validation** — asserts a stored internal/metadata URL (and bracketed-IPv6 `[::1]`) is blocked and never fetched. Also reordered the check to run *before* signing/fetching.
- **llm-feedback injection filtering** — asserts a prompt-injection feedback comment is dropped before the LLM call while benign feedback is kept.
- **OTLP depth cap** — asserts a 10k-deep nested attribute no longer risks a stack overflow during transform.
- **Production trace-ingestion key validation** — asserts prod refuses to start with a short/empty key when ingestion is enabled, and accepts a strong key / allows a weak key when disabled.

**Tradeoffs noted in review** (prompt-guard withholding false-positives; `acknowledge` now operator-gated) are intentional and left as-is for maintainer sign-off.

All affected suites green; build, typecheck, lint clean.

PR #1569, opened 2026-07-18
fix(#1567): exclude unknown CPU/memory samples from fleet aggregates

### LINKED ISSUE #1567: Metrics: Overall CPU and Memory usage under-reported due to 0% fallbacks
### Description
The overall CPU and Memory usage on the dashboard is reporting significantly lower values than expected. This is caused by the metrics collection logic defaulting to 0% when certain Docker API fields are missing.

### Technical Findings
The logic in `packages/observability/src/services/metrics-collector.ts` handles missing API fields by defaulting to 0, which skews the aggregated fleet metrics.

#### 1. CPU Calculation Issue
The CPU percentage is calculated as:
`cpuPercent = (cpuDelta / systemDelta) * numCpus * 100`

If `system_cpu_usage` is missing from either the current or previous snapshot, `systemDelta` becomes 0. The code then skips the calculation, leaving `cpuPercent = 0`.

#### 2. Memory Calculation Issue
The memory percentage is calculated as:
`memoryPercent = (memoryBytes / memoryLimit) * 100`

If the `memory_stats.limit` field is missing from the Portainer API response, `memoryLimit` defaults to 0, and `memoryPercent` remains 0.

### Impact
These 0% values are stored in the `metrics` table and used in overall fleet aggregations, resulting in a misleadingly low 'overall' usage metric.

### Proposed Fix
- Implement robust fallbacks for `systemDelta` and `memoryLimit`.
- If a metric cannot be reliably calculated, it should be marked as 'unknown' or omitted rather than reported as 0%.
- For memory, if the container limit is missing, consider using the total node memory as a fallback divisor.
- For CPU, handle cases where `systemDelta == 0` by avoiding the write or using a different calculation method.

laboef1900:
## Root cause

`collectMetrics()` (`packages/observability/src/services/metrics-collector.ts`) computed CPU% as `(cpuDelta / systemDelta) * numCpus * 100` and memory% as `(memoryBytes / memoryLimit) * 100`. When `system_cpu_usage` was missing from either Docker stats snapshot, `systemDelta` defaulted to `0` via `?? 0`, which skipped the calculation guard and left `cpuPercent` at its initialized `0`. Same story for memory: a missing `memory_stats.limit` defaulted to `0`, failed the `> 0` guard, and `memoryPercent` stayed `0`. Both fabricated zeros were then persisted to the `metrics` table and folded into the fleet-wide CPU/memory averages the dashboard reads — a container whose stats simply couldn't be computed this cycle was silently counted as "idle," dragging the whole fleet's reported usage down.

## Fix

- **`CollectedMetrics.cpu`/`memory` are now `number | null`.** `null` means "could not be reliably computed this cycle" (missing/invalid `system_cpu_usage` for CPU, missing/non-positive `memory_stats.limit` for memory) — never conflated with a genuine `0`.
- **The scheduler skips the write, not the schema.** `packages/server/src/scheduler.ts` no longer emits a `cpu`/`memory` row for a `metric_type` when the value is `null`. The `metrics.value` column stays `NOT NULL` — **no DB migration needed** — "unknown" is represented by the row's *absence* for that cycle, not a null cell. `AVG()` in PostgreSQL already skips rows it never sees, so the TimescaleDB continuous aggregates (`metrics_5min/1hour/1day`) and the SQL-side reports in `packages/observability/src/routes/reports.ts` get correct behavior for free.
- **The in-process fleet aggregation is null-safe.** `buildFleetResources()` (`packages/foundation/src/routes/dashboard.ts`, shared by `/api/dashboard/resources` and `/api/dashboard/full`) previously used one shared `statsCount` divisor and defaulted a missing `cpu`/`memory` key to `0` before summing — so a container with memory-only data still diluted the CPU average with a fake 0. It now tracks independent CPU and memory sample counts, both fleet-wide and per-stack, so a container missing one metric no longer drags down the other metric's average (or another container's).
- **Rejected: falling back to total node memory as the divisor** (the issue's suggested alternative for missing memory limits). Docker already reports the host's total memory as `memory_stats.limit` when a container has no explicit `--memory` cap — so a limit that's still missing/non-positive by the time it reaches our code signals a malformed or incomplete stats payload, not "no limit configured." Guessing a divisor for data already known to be untrustworthy would just trade one silent wrong number for another; a new node-memory lookup would also add a new external call/cache dependency for a case that's already rare. `null` (excluded from the aggregate) is the honest answer.

## Files changed

- `packages/observability/src/services/metrics-collector.ts` — `cpu`/`memory` become `number | null`; explicit presence checks instead of `?? 0`.
- `packages/observability/src/__tests__/metrics-collector.test.ts` — new tests: missing `system_cpu_usage` (current/previous snapshot), non-positive system-delta (counter reset), missing/zero `memory_stats.limit`, both unknown at once; precise numeric assertions on the happy path.
- `packages/server/src/scheduler.ts` — skips the `cpu`/`memory` metric-row write when the corresponding value is `null`.
- `packages/server/src/__tests__/scheduler.test.ts` — new tests asserting the skipped write (both/one unknown).
- `packages/foundation/src/routes/dashboard.ts` — `buildFleetResources()` tracks independent CPU/memory sample counts (fleet + per-stack) instead of one shared count with a `?? 0` default.
- `backend/src/routes/dashboard.test.ts` — **the key aggregate-level regression test**: a container with a known memory value but no `cpu` key no longer drags the fleet CPU average toward 0 (verified this fails on the pre-fix code: `25` instead of the expected `50`).
- `docs/architecture.md`, `CLAUDE.md` — new section documenting the unknown-vs-zero behavior and the rejected total-node-memory fallback.

No DB migration. No env var changes.

## Test plan

- [x] `npx vitest run` in `packages/observability` — 316 passed (full suite, DB-backed included)
- [x] `npx vitest run` in `packages/server` — 79 passed
- [x] `npx vitest run` in `packages/foundation` — 51 passed
- [x] `npx vitest run` in `backend` (integration suite against rebuilt `dist/`) — 471 passed, 1 skipped
- [x] `npx vitest run` in `packages/ai-intelligence` and `packages/operations` (consumers of `getLatestMetrics`/`getLatestMetricsBatch`) — 1063 / 225 passed, no regressions
- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] Manually reverted the `dashboard.ts` fix and re-ran the new aggregate test to confirm it fails on the old behavior (`25` vs expected `50`), then restored the fix and confirmed it passes.

Closes #1567


laboef1900 (commented):
## Automated PR Review

**PR**: #1569 -- fix(#1567): exclude unknown CPU/memory samples from fleet aggregates
**Branch**: `feature/1567-exclude-unknown-metric-samples` -> `dev`
**Diff size**: 603 lines (+385/-53)
**Reviewed at**: 2026-07-18

---

### Static Checks
| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | Root typecheck completed successfully |
| Lint | PASS | Backend/frontend ESLint and bash syntax checks passed |
| Backend Tests | PASS | 27 files; 471 passed, 1 skipped |
| Frontend Tests | PASS | 225 files; 2234 passed |
| Production Build | PASS | All workspaces built successfully |
| npm audit | FAIL | Current lockfile reports 7 high and 1 moderate production advisories; this PR changes no dependency files, so this is repository-wide baseline debt rather than a PR-introduced regression. The PR's GitHub Security Audit check is green. |
| Branch Target | PASS | `feature/*` correctly targets `dev` |

### Docker App Test
| Check | Status |
|-------|--------|
| Login | SKIPPED |
| Backend /health | SKIPPED |
| API /dashboard | SKIPPED |
| API /containers | SKIPPED |
| Frontend loads | SKIPPED |
| Migrations | N/A |
| No error logs | SKIPPED |

Docker/Playwright validation was skipped because required local port 5433 is already occupied; no shared development service or volume was disturbed. GitHub CI is green across backend, frontend, package tests, lint, typecheck, build, images, security audit, and the required test gate.

### Code Review Findings
| 1 | WARNING | `packages/observability/src/services/metrics-store.ts:304-336` | Omitting this cycle's CPU/memory row does **not** make the key disappear from `getLatestMetrics{Batch}`. Both queries independently select the newest row per metric type anywhere in the 15-minute window, so the previous cycle's CPU remains visible for up to 15 minutes. The new dashboard test mocks `{ memory, no cpu }`, but the real store cannot produce that immediately after one unknown CPU sample while an earlier CPU row is recent. Make latest reads cycle-coherent (return only rows sharing the container's newest collection timestamp) and cover that SQL contract. |
| 2 | WARNING | `packages/observability/src/services/metrics-collector.ts:79-89` | Missing `memory_stats.usage` is still coerced through `?? 0`; with a valid limit this reports a fabricated `0%`, recreating the exact unknown-as-idle bug for the numerator instead of the denominator. Require a present usage value before calculating memory percentage and add the missing-usage regression case. |
| 3 | WARNING | `packages/foundation/src/routes/dashboard.ts:157-160` | A latest-cycle record containing only `memory_bytes` (both percentages unknown) is discarded before `containerMetrics` is populated. The scheduler deliberately keeps writing known byte usage in that case, so per-stack `memoryBytes` now loses reliable data. Retain records when any of CPU, memory, or memory_bytes is present and test the per-stack total. |

### Acceptance Criteria (Issue #1567)
| AC-1 | Missing/invalid CPU denominator is unknown, not zero | PASS | Collector returns `null` for missing system counters and non-positive system delta |
| AC-2 | Missing/invalid memory inputs are unknown, not zero | PARTIAL | Missing/non-positive limit is handled, but missing usage still becomes 0% |
| AC-3 | Unknown percentage samples are omitted from persistence | PASS | Scheduler skips null CPU/memory rows and tests assert the written metric types |
| AC-4 | Fleet/per-stack aggregates exclude unknown values | PARTIAL | Independent counts are correct for an actually missing key, but latest-store reads carry forward the previous metric and raw memory bytes can be dropped |
| AC-5 | Tests and architecture guidance cover the behavior | PARTIAL | Strong unit coverage, but the dashboard mock does not model the real latest-row query semantics |

### Story Validation
**Issue intent**: Stop incomplete Docker stats payloads from being persisted and averaged as genuine 0% utilization.
**User journey**: Before: missing denominators silently lowered fleet averages -> Intended after: only reliable samples contribute. The collector/scheduler move in that direction, but the latest-read path still carries the previous metric forward and one missing numerator still produces 0%.
**Gaps**: Latest-cycle coherence, missing memory usage, and preservation of independently known memory-byte totals.

### Suggested Actions
1. **[SHOULD FIX]** Make `getLatestMetrics()` and `getLatestMetricsBatch()` return one coherent newest collection cycle per container.
2. **[SHOULD FIX]** Treat missing memory usage as unknown and cover it.
3. **[SHOULD FIX]** Preserve memory-byte-only records in stack aggregation.
4. **[REPOSITORY DEBT]** Triage current production dependency advisories separately; this PR does not alter dependencies.

### Verdict: COMMENT
**Reason**: The design is sound, but the real latest-metrics query prevents the key omission assumed by the aggregate regression test, and two remaining zero/drop paths leave the story only partially complete.

---
Reviewed by PR Reviewer Agent | 2026-07-18

laboef1900:
## Review fixes applied

Pushed commit `8fdd6ff1` to the PR branch:

- Made latest metric reads cycle-coherent so an omitted unknown metric is not carried forward from an older cycle.
- Treated missing `memory_stats.usage` as an unknown memory percentage rather than 0%.
- Preserved independently known `memory_bytes` when both percentage samples are absent.
- Added regression coverage and updated architecture guidance.

Verification after the fix:
- Typecheck: PASS
- Lint: PASS
- Observability package: 317 passed
- Foundation package: 51 passed
- Server package: 79 passed
- Backend: 472 passed, 1 skipped
- Frontend: 2234 passed
- Production build: PASS
- Pre-push typecheck + frontend suite: PASS

The DB-backed package suite used a disposable PostgreSQL container on port 55433, which was removed immediately after testing. Docker app/E2E validation remained skipped because shared port 5433 was occupied. Current production advisories are repository-wide and unchanged by this dependency-neutral PR.

PR #1571, opened 2026-07-18
fix(#1566): pin Edge heartbeat status to snapshot fetch time, not read time

### LINKED ISSUE #1566: Intermittent "All Hosts Down" status in UI due to SWR Cache/Heartbeat conflict
## Summary

The UI intermittently reports almost all docker-hosts as "down", flipping back to "up" every few minutes.

## Root Cause

In `packages/core/dist/portainer/portainer-normalizers.js` (within `determineEdgeStatus`):
```typescript
const elapsed = Math.floor(Date.now() / 1000) - lastCheckIn;
return elapsed <= generousThreshold ? 'up' : 'down';
```

The system uses a Stale-While-Revalidate (SWR) cache for the endpoints list with a TTL of 15 minutes (`TTL.ENDPOINTS = 900`). The Edge heartbeat threshold (`generousThreshold`) is approximately 90 seconds. 

When the monitoring service serves the endpoints list from the cache, it compares the cached `LastCheckInDate` with the current wall-clock time (`Date.now()`). If the cache entry is older than 90 seconds—which is frequent given the 15-minute TTL—the hosts are marked as `down` even if they are healthy. They only return to `up` once the SWR background process refreshes the cache.

## Steps to Reproduce

1. Monitor the `monitoring_snapshots` table in the database.
2. Observe that `endpoints_up` and `endpoints_down` flip between "all up" and "all down" states periodically.
3. Notice that the "all down" state corresponds with the period where the SWR cache is serving stale data.

## Expected Behavior

Endpoints should be marked as `up` if they were healthy at the time the snapshot was fetched, regardless of how long that snapshot has been in the cache.

## Actual Behavior

Endpoints are marked `down` based on the time elapsed since their last check-in relative to the *current* time, not the *snapshot* time.

## Fix Approach

1. Update `normalizeEndpoint` to include the timestamp of when the API response was received.
2. Pass this snapshot timestamp to `determineEdgeStatus`.
3. Replace `Date.now()` with the snapshot timestamp to calculate the elapsed time since the last check-in.

## Relevant Files

- `packages/core/dist/portainer/portainer-normalizers.js` — Heartbeat calculation logic.
- `packages/core/dist/portainer/portainer-cache.js` — SWR cache configuration (`TTL.ENDPOINTS`).

laboef1900:
## Root cause

`determineEdgeStatus()` compared an Edge endpoint's cached `LastCheckInDate` with `Date.now()`. The endpoints list is cached for 15 minutes, while the heartbeat threshold is about 90 seconds, so a healthy cached endpoint could flip to `down` until SWR refreshed the list.

## Fix

- **Cache snapshot metadata:** `fetchedAt` lives on the same L1 cache entry and L2 Redis envelope as its payload, so data and timestamp share serialization, expiry, eviction, and invalidation.
- **Atomic reads:** `cachedFetchSnapshot()` and `cachedFetchSWRSnapshot()` return `{ data, fetchedAt }` from the same selected cache entry. This prevents a background revalidation from pairing an already-returned stale payload with the replacement entry's timestamp. `getSnapshotTimestamp()` remains diagnostic/backwards-compatible only.
- **Heartbeat evaluation:** `normalizeEndpointAsOf(ep, { referenceTimeMs })` evaluates Edge heartbeat freshness against the snapshot's `fetchedAt`. `normalizeEndpoint(ep)` keeps its single-argument signature so existing bare `.map(normalizeEndpoint)` calls cannot mistake an array index for a timestamp.

## Covered consumers

All cached endpoint consumers whose behavior depends on `.status` now use the atomic snapshot pair:

- `packages/core/src/portainer/live-fleet.ts`
- `packages/foundation/src/routes/{dashboard,containers,endpoints,search,kubernetes,stacks,networks}.ts`
- `packages/ai-intelligence/src/routes/llm.ts`
- `packages/ai-intelligence/src/services/{monitoring-service,llm-tools}.ts`

Live/uncached callers and consumers that only read time-independent fields such as capabilities keep `normalizeEndpoint()`.

## Regression coverage

- Cache tests verify fresh snapshots, L2-to-L1 timestamp preservation, disabled-cache timestamps, and the completing-SWR race where revalidation lands before the caller resumes.
- Normalizer tests reproduce the stale-heartbeat failure and verify the explicit reference time plus `.map()` safety.
- Live-fleet coverage exercises the real SWR cache without leaking an unresolved revalidation.
- Dashboard and container route tests verify that a healthy Edge host remains usable after the cached endpoint list crosses the SWR stale boundary.
- Monitoring-service tests assert the atomic endpoint snapshot boundary.

## Verification

- `npm run typecheck` — passed
- `npm run lint` — passed
- `npm run build` — passed
- Full monorepo test suite against an isolated PostgreSQL test database — **5,812 passed, 1 skipped**
- GitHub CI was green before the review follow-up; a new run is triggered by the pushed fix commit

## Documentation

`docs/architecture.md` now describes the embedded L1/L2 metadata and atomic snapshot API rather than the superseded side-map design.

Closes #1566


laboef1900 (commented):
## Automated PR Review

**PR**: #1571 -- fix(#1566): pin Edge heartbeat status to snapshot fetch time, not read time
**Branch**: feature/1566-swr-heartbeat-snapshot-time -> dev
**Diff size**: 711 lines (+646/-65)
**Reviewed at**: 2026-07-18

---

### Static Checks
| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | `npm run typecheck` |
| Lint | PASS | `npm run lint` |
| Backend Tests | PASS | 470 passed, 1 skipped |
| Frontend Tests | PASS | 2234 passed |
| Production Build | PASS | `npm run build` |
| npm audit | FAIL (baseline) | 8 existing production findings (1 moderate, 7 high); no dependency changes in this PR, and GitHub Security Audit is green |
| Branch Target | PASS | Feature branch targets `dev` |

### Docker App Test
| Check | Status |
|-------|--------|
| Login | SKIPPED |
| Backend /health | SKIPPED |
| API /dashboard | SKIPPED |
| API /containers | SKIPPED |
| Frontend loads | SKIPPED |
| Migrations | N/A |
| No error logs | SKIPPED |

Docker and Playwright were skipped because the developer's existing stack owns required port 5433; it was not disturbed.

### Code Review Findings
| 1 | WARNING | `packages/foundation/src/routes/dashboard.ts:235-248,325-338,399-416`; `packages/foundation/src/routes/containers.ts:58-73` | The fix does not cover all status-dependent consumers of the cached endpoints list. `/api/dashboard/summary`, `/api/dashboard/resources`, and especially `/api/dashboard/full` (the home-page data source) still call `normalizeEndpoint` against the current wall clock, while `/api/containers` and `/api/containers/count` still filter cached endpoints the same way. Consequently, the reported UI can still show cached healthy hosts as down and the container views can temporarily drop their containers. A parallel “file lane” is not a safe reason to leave the primary user journey unfixed. |
| 2 | WARNING | `packages/core/src/portainer/portainer-cache.ts:831-832,955-989`; `packages/core/src/portainer/live-fleet.ts:96-105` | Payload and timestamp are not returned atomically. Each caller awaits `cachedFetch(SWR)` and then separately reads `getSnapshotTimestamp(key)`. A background revalidation can replace the cache entry in between, pairing the old payload already returned to the caller with the new entry's `fetchedAt`. I reproduced this with an immediately resolving SWR revalidation: the served value was `v1`, while `getSnapshotTimestamp` had already advanced to the `v2` fetch time. The new tests avoid this race by holding revalidation pending (and the live-fleet test leaves it unresolved forever), so they do not guard the actual boundary. |
| 3 | INFO | `docs/architecture.md:38`; PR description | The documentation and PR description say `getSnapshotTimestamp` uses a side-channel `Map`, but the current implementation deliberately embeds `fetchedAt` in the L1 entry and L2 envelope. This makes the architectural record misleading and contradicts the code/test comments. |

### Acceptance Criteria (Issue #1566)
| AC-1 | Cached healthy endpoints remain up relative to the snapshot fetch time | PARTIAL | Implemented in several core/foundation/AI paths, but the home dashboard and container routes remain current-time consumers. |
| AC-2 | Carry the response snapshot timestamp into heartbeat evaluation | PARTIAL | `fetchedAt` is stored with cache entries and `normalizeEndpointAsOf` accepts it, but callers retrieve data and time in two non-atomic operations. |
| AC-3 | Regression coverage reproduces and prevents the SWR/heartbeat conflict | PARTIAL | Core tests cover pending revalidation, but not a completing revalidation race or the primary dashboard/container routes. |

### Story Validation
**Issue intent**: Stop intermittent “All Hosts Down” UI flapping caused by comparing a cached Edge heartbeat with the current wall clock.
**User journey**: Before: a cached endpoint ages past the heartbeat threshold and appears down -> After: selected services pin evaluation to cache fetch time, but the home dashboard can still take the original broken path and a revalidation race can still mismatch data/time.
**Gaps**: Complete cached-endpoint caller coverage and make the returned snapshot payload/timestamp atomic.

### Suggested Actions
1. **[SHOULD FIX]** Add an atomic cache API that returns `{ data, fetchedAt }` from the same selected L1/L2/miss snapshot, including SWR stale serves, and migrate status-dependent endpoint consumers to it.
2. **[SHOULD FIX]** Wire the dashboard and container routes into snapshot-time normalization and add route-level regression coverage.
3. **[OPTIONAL]** Update architecture docs and the PR description to describe the embedded cache-entry/envelope design actually shipped.

### Verdict: COMMENT
**Reason**: Static checks and CI are green, but two logic gaps leave issue #1566 only partially addressed; Docker validation was also unavailable because the required port is in use.

---
Reviewed by PR Reviewer Agent | 2026-07-18

laboef1900:
## Review follow-up

All review findings were addressed in `da9d2f9b` and pushed to this PR branch:

- Added atomic `cachedFetchSnapshot()` / `cachedFetchSWRSnapshot()` results so a stale payload cannot be paired with a replacement entry's timestamp.
- Migrated every cached endpoint consumer whose behavior depends on status, including `/api/dashboard/{summary,resources,full}` and `/api/containers`/`count`.
- Added completing-revalidation, L2 timestamp, live-fleet cleanup, dashboard, container, and monitoring regression coverage.
- Updated the architecture documentation and PR description to reflect the embedded L1/L2 metadata design.

Verification after the fix:

- TypeScript: PASS
- Lint: PASS
- Production build: PASS
- Full monorepo tests against an isolated PostgreSQL database: **5,812 passed, 1 skipped**
- GitHub CI, package tests, frontend/backend tests, security audit, build, and Docker image builds: PASS
- Local Docker app/Playwright smoke test: SKIPPED because the developer's existing stack owns required port 5433; it was not disturbed

The PR is labeled **Reviewed by Agent** because the local Docker smoke phase was unavailable; no code-review findings remain.

Issue #1587, opened 2026-07-19
frontend has no boundary enforcement on @dashboard/* imports


laboef1900:
## Summary

#1585 added architectural boundary enforcement for `packages/`, but `frontend/` is outside that
config's scope. Nothing prevents a frontend file from importing `@dashboard/core`, a domain package,
or `@dashboard/server` — which would pull backend code, and its transitive dependency fan-out, into
the browser bundle.

Deliberately deferred from #1585 so that PR's red/green signal stayed about `packages/`.

## Root Cause

`eslint.packages.config.mjs` scopes itself to the code it polices:

```js
files: ['packages/*/src/**/*.ts', 'packages/*/scripts/**/*.ts'],
```

`frontend/eslint.config.js` has no `eslint-plugin-boundaries` section at all, so no rule evaluates
any frontend import.

Today's usage is correct and narrow — 6 sites, all `import type`, all `@dashboard/contracts`:

- `frontend/src/features/ai-intelligence/hooks/use-investigations.ts`
- `frontend/src/features/ai-intelligence/hooks/use-monitoring.ts`
- `frontend/src/features/ai-intelligence/hooks/use-incidents.ts`
- `frontend/src/features/ai-intelligence/hooks/use-incident-insights.ts`
- `frontend/src/features/ai-intelligence/components/insight-card.tsx`
- `frontend/src/features/ai-intelligence/components/insight-card.test.tsx`

The issue is that this correctness rests on convention, which is exactly the state `packages/` was in
before #1585.

## Steps to Reproduce

1. Add `import { getAppDb } from '@dashboard/core/db/app-db.js';` to any file under `frontend/src/`
2. `npm run lint`
3. Observe: exit 0

## Expected Behavior

A frontend file importing anything other than `@dashboard/contracts` fails lint. Contracts is the
correct shared surface: it is types and Zod schemas only, with no runtime backend dependency.

## Actual Behavior

Lint exits 0. No rule evaluates any frontend import.

## Fix Approach

1. Add an `eslint-plugin-boundaries` section to `frontend/eslint.config.js` with two elements
   (`frontend`, `contracts`) and a single policy allowing `frontend -> contracts` only
2. Reuse the `tsconfig.eslint.json` path-map approach from #1585 so `@dashboard/*` resolves to
   package source rather than gitignored `dist/` — a dist-resolved config passes vacuously on a
[...]

Issue #1586, opened 2026-07-19
Test files are excluded from every tsconfig.build.json, hiding 49 type errors from npm run typecheck


laboef1900:
## Summary

`npm run typecheck` never typechecks a single test file in `packages/`. All nine
`packages/*/tsconfig.build.json` files exclude `**/*.test.ts` and `**/__tests__/**`, and the root
`typecheck` script builds only that project. 49 type errors across 5 packages are currently invisible.

Found while implementing #1585, which fixed the sibling problem for lint: a gate whose scope silently
stopped covering the code it was written to protect.

## Root Cause

Root `package.json`:

```json
"typecheck": "tsc --build tsconfig.build.json && npm run typecheck -w frontend"
```

Every package's `tsconfig.build.json` (e.g. `packages/operations/tsconfig.build.json`):

```json
"exclude": ["**/*.test.ts", "**/__tests__/**", "node_modules", "dist"]
```

Excluding tests from the *build* is correct — they must not ship to `dist/`. The gap is that no
*other* project ever typechecks them, so `tsc -p tsconfig.json` (which does include tests) is never
run by any script or CI job.

## Steps to Reproduce

1. `npm run typecheck` — exits 0
2. `cd packages/core && npx tsc -p tsconfig.json --noEmit` — 24 errors
3. Repeat per package

## Expected Behavior

A type error in a test file fails `npm run typecheck`.

## Actual Behavior

`npm run typecheck` exits 0. Measured per package:

| Package | Errors when tests included |
|---------|---------------------------|
| `core` | 24 |
| `ai-intelligence` | 14 |
| `security` | 6 |
| `operations` | 4 |
| `foundation` | 1 |
| `contracts` / `infrastructure` / `observability` / `server` | 0 |

Representative failures:

`packages/operations/src/__tests__/remediation-service.test.ts` lines 93 and 389 — `MetricsInterface`
gained five members (`getLatestMetricsBatch`, `getMovingAverage`, `getCapacityForecasts`,
`generateForecast`, `findSimilarInsights`) and the mock was never updated:

```
error TS2345: Argument of type '{ getLatestMetrics: Mock; getMetrics: Mock; detectAnomalies: Mock; }'
is not assignable to parameter of type 'MetricsInterface'.
```

`packages/core/src/portainer/portainer-client.test.ts` lines 291-315 — global `Response` vs undici's
`Response`:

```
error TS2345: Argument of type 'Response' is not assignable to parameter of type
[...]

Issue #1578, opened 2026-07-18
ci: Security Audit job cannot fail on high-severity production advisories


laboef1900:
## Problem

The CI job named **Security Audit** cannot fail on High-severity production vulnerabilities. `.github/workflows/ci.yml`:

```yaml
- name: Audit production dependencies (high+)
  run: npm audit --omit=dev --audit-level=high
  continue-on-error: true          # <- never fails the job

- name: Audit all dependencies (critical gate)
  run: npm audit --audit-level=critical
```

The step matching the repo's documented `audit:prod` gate (`package.json`: `npm audit --omit=dev --audit-level=high`) is advisory-only. The sole enforcing step is `--audit-level=critical`. Since nothing currently rates critical, the job is green while `npm run audit:prod` exits 1 locally.

**A gate that cannot fail is not a gate.** It reports green and reads as coverage.

## Impact, concretely

This is why `dev` shipped **undici 8.3.0 carrying three High advisories** (CVE-2026-12151, CVE-2026-9697, CVE-2026-9675) with CI green throughout, until they were found by hand reviewing #1573. The same gap hid nodemailer's GHSA-p6gq-j5cr-w38f (High, CVSS 7.1) — see #1574 / #1576 — and currently hides the ws cluster in #1577.

Three separate High advisories reached the default branch unflagged. That is the failure mode, not a hypothetical.

## Sequencing — this cannot land first

Removing `continue-on-error` today turns CI red immediately: `npm audit --omit=dev` currently reports 5 highs. Order matters:

1. #1576 — nodemailer -> 9.0.3 *(open, awaiting merge)*
2. #1577 — ws cluster via transitive refresh *(4 of the 5 highs)*
3. #1575 — hono -> 4.12.30 *(the 5th)*
4. **Then** remove `continue-on-error: true` from the high+ prod step

Encouragingly, every current high reports `fixAvailable: true` — all are transitive refreshes, none require a breaking change. Reaching a clean prod audit is realistic, not aspirational.

## Design question for whoever picks this up

A bare `--audit-level=high` gate will eventually block an unrelated PR on a freshly-published advisory with no fix available, which is how gates get disabled again. Worth deciding up front between:

- **Allowlist of known-unfixable advisory IDs**, reviewed on a schedule — fails on anything *new*, which is the actual goal
- **Split jobs** — a blocking gate on `--omit=dev`, plus a separate non-blocking informational audit for dev deps
[...]

PR #1588, opened 2026-07-19
fix(lint): enforce packages/ architectural boundaries (#1585)

### LINKED ISSUE #1585: Architectural boundary rules enforce nothing since the packages/ restructure
## Summary

The `eslint-plugin-boundaries` rules in `backend/eslint.config.js` currently police zero files, and `packages/` is not linted at all. The dependency-direction guarantees documented in `packages/core/src/CLAUDE.md` are running on convention alone — nothing in CI would catch a violation.

Found while verifying the `eslint-plugin-boundaries` 6→7 bump in #1584 (checking that the rules still *catch* violations, since a silently-disabled rule passes lint exactly like a satisfied one).

## Root Cause

The element patterns still describe the pre-monorepo layout. In `backend/eslint.config.js` lines 30-37:

```javascript
'boundaries/elements': [
  { type: 'core', pattern: 'src/core/**', mode: 'folder' },
  { type: 'scheduler', pattern: 'src/scheduler/**', mode: 'folder' },
  { type: 'app', pattern: 'src/app.ts', mode: 'file' },
  { type: 'routes', pattern: 'src/routes/**', mode: 'folder' },
  { type: 'sockets', pattern: 'src/sockets/**', mode: 'folder' },
],
'boundaries/ignore': ['**/*.test.ts', '**/*.spec.ts', '**/test-utils/**'],
```

Every one of those paths moved to `packages/` during the restructure. What remains in `backend/src` is 37 TypeScript files: 29 tests, plus 8 under `test-utils/` and `test/` — and `boundaries/ignore` excludes all of them.

Two compounding gaps:

| Gap | Detail |
| --- | --- |
| Patterns are stale | `src/core/**`, `src/scheduler/**`, `src/app.ts`, `src/routes/**`, `src/sockets/**` match nothing; `backend/src/core/` does not exist |
| `packages/` is unlinted | Only `frontend/eslint.config.js` and `backend/eslint.config.js` exist. Root `lint` runs `npm run lint -w backend && npm run lint -w frontend`; `backend`'s script is `eslint --max-warnings=0 src/` |

So the code that the rules were written to protect — the `core/` kernel and the 9 domain packages — is outside every configured lint scope.

## Steps to Reproduce

1. `find backend/src -name '*.ts' ! -name '*.test.ts' ! -name '*.spec.ts'` — returns 8 files, all under `test-utils/` or `test/`
2. `find . -maxdepth 3 -name 'eslint.config.*' ! -path '*/node_modules/*'` — returns only `frontend/` and `backend/`
3. Add an import from a domain package into `packages/core/src/`, violating the documented direction
4. Run `npm run lint`
5. Observe: exit 0

## Expected Behavior

A `packages/core` file importing from a domain package, from `routes/`, or from `sockets/` fails lint, per the dependency direction in `packages/core/src/CLAUDE.md`.

## Actual Behavior

Lint exits 0. No rule evaluates any file under `packages/`.

## Fix Approach

1. Add an eslint config covering `packages/`, with `boundaries/elements` describing the real layout: `contracts` → `core` → the five domain packages → `foundation` → `server`
2. Encode the documented rules: domain packages may import only `core` + `contracts`; no domain package may import `foundation`; `@dashboard/ai` imports only core + contracts; routes are not re-exported from barrels
3. Extend root `lint` to cover the new scope (`packages/` is not currently a lint target in any script)
4. Prove the rules bite — add a fixture that violates each direction and assert eslint reports it, so this cannot silently regress a second time
5. While in the file: migrate the deprecated `mode` descriptors to `partialMatch` (v7 warns on them). Deliberately excluded from #1584 because it changes which files each element matches

## Relevant Files

- `backend/eslint.config.js` — stale element patterns (lines 30-37), ignore list (line 37)
- `packages/core/src/CLAUDE.md` — documents the dependency direction that should be enforced
- `package.json` — root `lint` script omits `packages/`
- `backend/package.json` line 11 — `"lint": "eslint --max-warnings=0 src/"`

**Effort Estimate:** Medium
**Impact Estimate:** High

laboef1900:
Closes #1585

## Problem

The `eslint-plugin-boundaries` rules in `backend/eslint.config.js` policed **zero files**. Every
element pattern (`src/core/**`, `src/routes/**`, `src/sockets/**`, `src/scheduler/**`, `src/app.ts`)
described the pre-monorepo layout; that code all moved to `packages/`. What remained in `backend/src`
was 37 files, all tests or test-utils, and `boundaries/ignore` excluded every one. Separately,
`packages/` was not linted at all.

The dependency directions in `packages/core/src/CLAUDE.md` were running on convention alone.

## The load-bearing design decision

`@dashboard/*` normally resolves through each package's `exports` map into `dist/`. **`dist/` is
gitignored**, so a dist-resolved config would pass vacuously on a fresh clone or a pre-build CI run —
recreating the exact bug this PR fixes, in a new file.

Resolution is therefore pinned to source via a lint-only `tsconfig.eslint.json`. Measured, with the
path map removed and a real violation planted:

| Import shape | In `exports` map | Resolves to | Result |
|---|---|---|---|
| `@dashboard/observability` | yes | `dist/index.d.ts` (unknown element) | exit 1, wrong message but loud |
| `@dashboard/observability/services/metrics-collector.js` | no | unresolvable, classified `external` | **exit 0, zero bytes** |

The second case is the one that silently rots, and it covers every deep subpath import — which is how
`core` is imported 609 times in this repo.

Verified end-to-end: with **all nine `dist/` directories moved aside**, a clean tree still exits 0 and
a planted violation is still caught.

## The rule table

Encodes the real graph, not the top-level diagram. `infrastructure` is a sanctioned **sub-tier**, not
a domain peer — `security` and `operations` import it (`assertCapability`, `getElasticsearchConfig`),
documented as Phase 3 exceptions in each package's own `CLAUDE.md`.

| From | May import |
|---|---|
| `contracts` | (nothing) |
| `core` | contracts |
| `infrastructure` | core, contracts |
| `observability` | core, contracts |
| `ai` | core, contracts (hard isolation) |
| `security` | infrastructure, core, contracts |
| `operations` | infrastructure, core, contracts |
| `foundation` | ai, observability, security, infrastructure, core, contracts — **never operations** |
| `server` | everything |

Plus: a package barrel (`src/index.ts`) must not re-export from `routes/`.

## Proving the rules bite

`backend/src/packages-boundaries.test.ts` (23 assertions) drives the **shipping config**, not a copy.
It covers each forbidden direction, a positive control (so the suite cannot pass by reporting
everything), build-state independence, and anti-decay guards asserting the weakening mechanisms are
absent — including that root `lint` still calls `lint:packages`, the assertion that would have caught
the original regression.

Verified by mutation — every weakening turns the suite red:

| Mutation | Result |
|---|---|
| root lint drops `&& npm run lint:packages` | 1 failed |
| `default: 'disallow'` → `'allow'` | 7 failed |
| drop resolver `project` | 9 failed |
| `checkInternals: true` → `false` | 2 failed |
| remove `no-unknown-files` | 2 failed |
| `partialMatch: false` → `true` | 1 failed |

## Also in this PR

- Adopts `js.configs.recommended` on `packages/` for parity with the other two workspaces, fixing the
  4 errors it surfaces. The `paramIdx` dead-store removals in `reports.ts` drop only a discarded
  return value — `addLifecycleRunningFilter` still mutates `conditions`/`params` by reference — and a
  new test asserts the SQL placeholder invariant that would break if that were wrong.
- Deletes the dead boundaries block from `backend/eslint.config.js`. `backend` is a test-only
  workspace that legitimately imports all 8 packages; there is no boundary left to police there.
- Removes 3 doc entries claiming cross-domain imports that do not exist (resolved via DI, not import
  edges). Verified no import statement backs them — `@dashboard/ai` is not even in `security`'s
  dependencies.
- Redraws the tier diagram in `packages/core/src/CLAUDE.md` and `docs/ai-instructions/architecture.md`,
  which contradicted the sanctioned exceptions.

## Verification

| Check | Result |
|---|---|
| `npm run lint` | exit 0 |
| `npm run typecheck` | exit 0 |
| `npm test` | 225 files, **2234 passed** |
| `packages-boundaries.test.ts` | 23/23 |
| Clean-checkout attack (all `dist/` removed) | violation still caught |

CI needs no workflow change — its lint job is a bare `npm run lint`.

## Notes for review

- The config is renamed `.mjs` rather than adding `"type": "module"` to root `package.json`, which
  would flip module semantics for every root-level `.js` in the repo. This also silences a
  `MODULE_TYPELESS_PACKAGE_JSON` warning that was appearing in every CI lint log.
- Four devDeps are now declared at root (`@eslint/js`, `typescript-eslint`, `eslint-plugin-boundaries`,
  `eslint-import-resolver-typescript`). They previously resolved only by accidental hoisting from
  `backend/`; `@eslint/js` was not resolvable from root at all.
- **Please do not trim the anti-decay assertions as redundant.** Removing any one of the resolver
  setting, `checkInternals`, or `no-unknown-files` makes the gate enforce nothing while still exiting
  0. Those assertions are the compensating control.
- Known gap, stated rather than implied: a red Lint job only blocks a merge if `Lint` is a **required
  status check** in branch protection, which cannot be verified from repo files.

Follow-ups filed while working on this: #1586 (test files excluded from every `tsconfig.build.json`,
hiding 49 type errors), #1587 (frontend has no boundary enforcement).


laboef1900:
## Review

A boundary gate's failure mode is *vacuous pass* — the config loads, lints, and exits 0 while enforcing nothing. Reading the diff proves very little about that, so I re-ran the claims independently rather than grading the tables in the description.

### Verification

| Check | Result |
|---|---|
| `npm run lint` (full root chain) | exit 0, reaches `lint:packages` |
| `packages-boundaries.test.ts` | 23/23 |
| `reports-route.test.ts` | 30/30 |
| All 6 mutations in the PR's table | Reproduce with **exactly** the claimed counts (9/2/7/2/1/1) |
| Barrel policy moved off last position | 1 failed (also guarded) |
| Deep-subpath violation planted in `packages/core` | Caught, correct `core -> observability` message |
| Same, with `packages/observability/dist` **moved aside** | Still caught — build-independence confirmed |
| New package dir with no `paths`/element entry | 2 failed |
| Off-by-one planted in the lifecycle placeholder index | New SQL test catches it |

I also probed three bypasses the PR does not claim to cover. All three are caught:

- Cross-package import via **relative path** (`../../observability/src/...`), sidestepping the `@dashboard/*` path map entirely — reported with the correct element types.
- A package importing **out of `packages/` into `backend/`** — `no-unknown-dependencies`.
- Same into `frontend/` — `no-unknown-dependencies`.

A caveat on method, since it would have produced three false findings: my first mutation run showed `checkInternals`, `partialMatch`, and the barrel-position mutations as *surviving*. That was my error — the search-and-replace hit the first occurrence, which lives in the header comment. Re-targeted at the code, all three fail as documented.

The "4 errors surfaced by `js.configs.recommended`" claim also checks out — reverting the six source edits yields exactly 2x `no-useless-assignment` and 2x `preserve-caught-error`, plus 6 unused-disable-directive warnings that `--max-warnings=0` would fail on.

### Correctness

**The `paramIdx` dead-store removal is safe.** `addLifecycleRunningFilter` is called at `reports.ts:470` and `:595`, and in both blocks it is the last statement before `conditions.join(' AND ')`. `paramIdx` is scoped to the `withStatementTimeout` closure with no reference after either call — confirmed against all 28 occurrences in the file, not just the diff hunks.

**The `{ cause: err }` additions cannot leak filesystem paths.** Worth checking, because `PathTraversalError` carries public `baseDir` and `segment` fields and the security rules require stripping filesystem paths before they reach the frontend. Both call paths catch and return a generic body (`backup.ts` -> `{ error: 'Invalid backup filename' }`, `portainer-backup.ts:87-91` -> `{ error: 'Invalid filename' }`, both 400). `error-handler.ts` never touches `.cause`, and a cause set via the options bag is non-enumerable so it does not serialize. Server-side log enrichment only.

### Findings

**1. `tsconfig.eslint.json:38-40` is factually wrong.** It says a missing `paths` entry "is NOT caught by anything." It is caught — `covers every directory under packages/ in the path map` diffs the map against a live `readdir`. I confirmed with a throwaway `packages/zzprobe/src/x.ts`: two failures, that assertion plus the element-descriptor one. Given this PR's thesis is that documentation claiming enforcement it doesn't have is the bug, the inverse error is worth fixing — it understates the safety net and invites someone to rebuild machinery that already exists.

**2. `addLifecycleRunningFilter`'s return type is now dead.** Both call sites discard it, so `: number` describes a contract nobody honors. `void` makes "this must be called last" structural rather than a convention the next person re-derives. As written, someone appending a filter below the call site would reuse the placeholder it just spent — the new test would catch that on the endpoint-scoped path, but the signature is what invited the mistake.

**3. `architecture.md`'s diagram still implies the tiering its own prose has to correct.** The redraw separates `ai, observability` from `security, operations` with a blank line, but leaves them stacked above, so the paragraph below has to explain that they "sit alongside `security`/`operations`, not above them." It also drops the `foundation -> security` arrow, which is an allowed edge.

**4. `architecture.md` contains a claim the code contradicts.** "nothing in the tree depends on `ai`" is false — both `foundation` and `server` declare `@dashboard/ai` as a dependency and import from it. `ai`'s hard isolation is one-directional: it constrains what LLM code may reach, not who may reach it. `packages/core/src/CLAUDE.md` phrases this correctly already, so the error is confined to that one sentence. Flagging it because it is the same failure class this PR exists to kill, in the PR's own prose.

### Risks

Low. Nothing changes runtime behaviour except two error objects gaining a `cause` and two provably dead stores. The gate is additive and CI needs no workflow change (`ci.yml:96` is a bare `npm run lint`).

The author's stated gap is the real one: a red `Lint` job only blocks a merge if `Lint` is a **required status check** in branch protection. That is unverifiable from repo files and is `continue-on-error` by another name — the same shape as #1578. Worth confirming in repo settings before treating this as closed.

Second-order: the four devDeps moved to root previously resolved by accidental hoisting from `backend/`. `backend/package.json` correctly still declares `@eslint/js` and `typescript-eslint` for its own config, and root now declares all four — but root and `backend` must now keep their eslint versions compatible.

### Verdict

**Approve.** The anti-decay assertions are the load-bearing part of this PR and they survived every mutation I could construct, including three the PR doesn't claim. I'd endorse the request not to trim them.

All four findings are non-blocking. A follow-up commit addressing them is prepared on this branch (`docs(lint): correct three inaccuracies in the boundary-gate docs`), verified with `npm run lint` (0), `npm run typecheck` (0), `packages-boundaries.test.ts` (23/23), the full `packages/observability` suite (29 files, 323 passed) and the full `backend` suite (30 files, 516 passed).

PR #1616, opened 2026-07-27
fix: act on the round-2 design critique — the prose was authored, the numbers were not

laboef1900:
Acts on a second whole-application design review, run specifically to hunt "AI slop".

The finding worth leading with: **the prose is already good and the numbers are not.** Almost every
defect had one shape — the system computes a qualifier (`status='error'`, `r_squared`, `confidence`,
a sample count, which table a value came from) and then **discards it at the render boundary** while
keeping the figure it qualifies. That is where the slop lives in this product, not in the
typography, and in a monitoring tool it is the only place it really matters.

Every headline finding was re-verified against source or the live API before it was fixed, and each
regression guard was re-run with the defect reintroduced to prove it actually fails.

## P0 — the dashboard could not count a stopped container

`health-score.ts` compared against `'exited'`. That is Docker's word; `normalizeContainer` maps it
to `'stopped'` before any client sees it, so the branch was unreachable and `stats.stopped` was
pinned at 0. Home and Health & Monitoring both render that number, so both reported
**"Needs attention 0 · No unhealthy or stopped containers"** under a green check — while the live
API said `stopped: 6` and the same page drew red Stopped bars 350px lower.

It survived because **every fixture in `health-score.test.ts` was built with `state: 'exited'`**.
The suite agreed with the bug. A test suite that invents its own vocabulary cannot catch a
vocabulary error.

`CONTAINER_STATES` now lives once in `@dashboard/contracts`, is serialized through
`ContainerStateSchema`, and is derived from by both the normalizer and the frontend's `Container`
type — so a Docker-native word at a call site is a compile error. Same root cause fixed in the dead
`?state=exited` drilldown, the phantom "Exited" filter option, the correlated-anomaly filter, and
the comparison view. `dead` was counted nowhere and now is.

## P0 — the render-boundary class (8 sites)

- **`Error Rate: 10000.0%`** — `getLlmStats` returns a percentage; the tile multiplied by 100 again.
  The threshold check compared `> 0.05` against that percentage, so "above the 5% threshold" fired
  at 0.05%.
- **`p95` above `max`, on four live rows.** Aggregates come from a rollup above 6h — a continuous
  aggregate that refreshes on a policy, so it lags — while percentiles can only come from raw
  `metrics`. Two populations printed as one row. Percentiles now render only when the aggregates
  also read raw metrics; otherwise `null`, with the reason stated.
- **`percentile_cont` over an empty set returns NULL**, which `Number()` turned into a confident
  `p95 0.00%` for containers that reported nothing.
- **Errored LLM calls folded into every aggregate** — one call that failed before it left the
  process produced "Avg Latency 69ms" and credited `gpt-4o-mini` with 100% share of a request it
  never received.
- **A `~1h` breach ETA for a container at 0.0% CPU**, from an OLS fit with no goodness-of-fit gate.
  The backend already returns `confidence` and `r_squared`; the table rendered neither.
- **The "Risk-ranked" table inverted its own severity column** — incompatible scales meant every
  breach more than four hours out ranked below rows the same table labelled Healthy.
- **Right-sizing advice off two samples**, beside containers with 189.
- **`13.8 GB total disk usage`** — a plain sum of per-image sizes, so every shared layer is counted
  once per image. `alpine:latest 13 MB` and `alpine:3.22 12.8 MB` are one blob billed as 25.8 MB.
- **`All 200 traces: … container: unknown`** while 3982 matched.

## P0 — one detector, two contradictory labels, one screen

`signature-meta.ts` mapped `ml-anomaly` to **"ML"**; `insight-card.tsx` mapped the same key to
**"Metric anomaly"**, under a twelve-line comment explaining exactly why it must not say ML. Both
rendered on `/health` at once, over rows describing themselves as `method: adaptive` — a moving
average with a threshold.

The reasoning had been written down and was lost anyway. It is now one module with a test that
fails if any consumer disagrees. Two further keys had already drifted the same way.

Also gone: `confidence: 1.00` on ~40 anomalies (a clamped restatement of the z-score),
`z-score: 1116.00` beside `mean: 0.0%` (a collapsed denominator, not a distance), and two
boilerplate sentences repeated verbatim on every card of their kind.

## P1/P2

- **Contrast** failed AA on **three** light themes, not the one a browser scan found.
  `--color-muted-foreground` is the most-used text colour in the app (325 elements on `/health`).
  `theme-contrast.test.ts` now holds every theme to 4.5:1 by parsing the shipped CSS.
- **Keyboard access** broke exactly where work happens: no skip link (22–29 tabs to content),
  `/reports` sort headers were `<span onClick>` with `aria-sort: null` and five of eight inert, the
  topology graph was an unnamed `role="application"` with ~70 unnamed stops, `/assistant` had zero
  `aria-live`, and the `/images` dialog never took focus.
- **Enumerated blocks distilled**: the security audit's phantom "20 containers need a look" (it
  counted rows, so expanding the clean list manufactured findings), 128 identical cells across four
  constant remediation columns, 15 uniform red badges that ranked "never enable this in production"
  level with a typo warning, nine label-restating help texts, and `Satisfaction Rate 0%` from zero
  samples.
- **Unconfigured features presented working controls**: `Start Capture` was live with
  `PCAP_ENABLED=false`, and `/assistant` offered four suggested questions with no reachable LLM.
- **Three logo pickers → one.** 30 radio cards to choose a logo three times became 11 controls.
  `faviconIcon` is the primary and overrides are derived, so there is **no new persisted key** —
  which matters, because this store has no `migrate` and zustand 5 discards the entire blob on an
  unmigrated version bump.
- **A guess is labelled a guess.** The log level column was a keyword grep presented as the
  emitter's own level — `module: "trace-store"` badged DEBUG for containing "trace", on the field
  operators triage on.

## Two reported findings deliberately **not** acted on

Both were checked and did not hold:

1. **`/edge-logs` conflating a 503 outage with "not configured"** — the route only 503s when
   `getElasticsearchConfig()` is absent; a failure against a *configured* cluster returns 502, which
   the page already renders as an outage with Retry.
2. **`no errors found` badged ERROR** — the keyword regex is word-bounded and does not match the
   plural. The singular does, and is covered by a test.

## Testing

frontend 2855 · core 989 · ai-intelligence 1112 · observability 342 · security 329 · operations 237
· backend 541 · contracts 31 · foundation 51 — all passing. `npm run typecheck` and `npm run lint`
clean. No dependency changes, so the audit and `npm ls --all` gates are unaffected.

New guards, each verified to fail against the original defect:
`container-state-vocabulary.test.ts`, `detection-method-labels.test.ts`, `theme-contrast.test.ts`,
`llm-stats-aggregation.test.ts` (real PostgreSQL), `forecast-ranking.test.ts`,
`log-level-source.test.ts`, `skip-link.test.tsx`, `app-icon-section.test.tsx`,
`anomaly-description.test.ts`.

## Docs

`docs/architecture.md` gains a "Design-critique remediation, round 2" section; `CLAUDE.md` gains
invariants 7–11. Invariant 8 is the meta-lesson and the one most worth keeping:

> **Record an honesty rule as a test, not a comment.** `insight-card.tsx` carried twelve lines
> explaining why a badge must not say "ML", and the rule was lost anyway because a sibling module
> kept its own copy of the map.

No new environment variables, so `docker/.env.example` is unchanged. `AGENTS.md`/`GEMINI.md` are
pointer files to `CLAUDE.md`, not mirrors.

## Note for the reviewer

Two API responses gained fields (`fleetSummary.totalObserved`, `aggregateSource` on
`/api/reports/utilization`; `failedQueries`/`succeededQueries` on the LLM stats payload) and two
read-only endpoints are new (`GET /api/pcap/status`, `GET /api/llm/status`, both `authenticate`-only
and reporting only whether a feature is on). Percentiles on `/api/reports/utilization` are now
`null` on ranges above 6h — a deliberate behaviour change, since the alternative was continuing to
print a 95th percentile above its own maximum.

https://claude.ai/code/session_01Y5FoXKQ1d7hJ3WjGfFUbQ5


laboef1900:
Remediation of the review findings pushed as 9 commits (`46734da2`..`ef1ea402`), taking this from 89 to 100 files.

Headline fix: `ReportsQuerySchema` capped `timeRange` at 24h/7d/30d while `selectRollupTable` only reads raw `metrics` at <= 6h, so the percentile branch this PR added was unreachable — p50/p95/p99 were always null and both p95-keyed right-sizing rules could never fire. The route tests hid it by mocking `selectRollupTable` to `isRollup: false`, including for a 7d request. A 6h range makes the branch reachable and the mock is gone.

Also: a real 30s TTL behind `GET /api/llm/status` (the route comment promised a cache that did not exist), two regression guards rewritten to drive the shipped code rather than local copies of it, `emittedLevel` tightened so prose no longer resolves as a stated level, two contrast failures on the log console, and a pass removing comments that asserted mechanisms the code does not have.

Verified locally: typecheck clean, lint clean, 6617 tests passing across frontend, all 9 packages and backend.

Merging with `Security Audit` red. It flags `react-router` GHSA-qwww-vcr4-c8h2, an **RSC-mode** CSRF bypass; this app uses `createBrowserRouter` + `RouterProvider` with no RSC usage, so it is not reachable here. There is no in-place upgrade (`react-router-dom@8` does not exist — v8 folded the DOM bindings into `react-router`), so clearing it is a 98-file import migration. Deferred to #1618. The other two production Highs (`@fastify/static`, GHSA-8pvw-jcv7-9cmj and GHSA-83w8-p2f5-377r) **are** fixed here via a targeted override.

Follow-ups filed: #1617 (frontend test files are excluded from typecheck), #1618 (react-router migration).

